PR #357, opened 2014-04-08
Allow raw query strings to be added to requests

dpolivy:
In some scenarios[1], it may be valuable to be able to pass
raw/pre-encoded query string parameters with a request. A recent change
to superagent forced all qs params to be utf8 encoded, which changed the
previous behavior of superagent. This commit adds another function,
`queryRaw`, to allow adding query string parameters that are NOT
re-encoded in any way prior to being passed as part of the request.

I thought it best to have an explicit approach to doing this to make
sure the developer intended this behavior, rather than trying to deduce
which querystring values should be encoded via other means.

[1] One example: when testing a server that uses a non-UTF8 URL encoding

See discussion on this commit for more context: https://github.com/visionmedia/superagent/commit/629bf9b10d26e2e6b44aea11b197ef8fea99a9c5

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -126,6 +126,7 @@ function Request(method, url) {
   this.attachments = [];
   this.cookies = '';
   this.qs = {};
+  this.qsRaw = [];
   this._redirectList = [];
   this.on('end', this.clearTimeout.bind(this));
   this.on('response', function(res){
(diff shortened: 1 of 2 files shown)


dpolivy:
Sure, I'm happy to revisit the API to use a parameter instead of a separate method. I don't have any strong preferences here -- are you OK if I revamp this to match what you proposed?

Also, not sure I understand the client concern you're asking about?

defunctzombie:
@dpolivy The client library and the server library need to support this feature.

defunctzombie:
Thinking about it more, is there a downside to detecting when a string is specified and just using that as the query? That way we don't need an options argument.

defunctzombie:
This PR doesn't work. the query args are encoded someplace else as well.

defunctzombie:
I will make a new PR for this for server and client versions.

RamIdeas:
Being able to accept a given query string as is would be helpful.

Please see #581

PR #366, opened 2014-04-23
make agent setcookie callback before end callback

alsotang:


defunctzombie:
Why was using

```
self._callback = function(){
  fn && process.nextTick(fn);
};
```

deemed worse than this listener re-arrangement?

Issue #298, opened 2013-12-07
pipe doesn't work when there is a redirect


esnunes:
I'm trying to download google profile picture using superagent in node. When you request a url like this: "https://profiles.google.com/s2/photos/profile/GOOGLE_ID?access_token=ACCESS_TOKEN", google does redirect with the following message on body:

```
    <HTML>
    <HEAD>
    <TITLE>Moved Temporarily</TITLE>
    </HEAD>
    <BODY BGCOLOR="#FFFFFF" TEXT="#000000">
    <H1>Moved Temporarily</H1>
    The document has moved <A HREF="https://lh6.googleusercontent.com/-e8DUYl6qAOM/AAAAAAAAAAI/AAAAAAAAAFI/5qRGDU7oU3g/photo.jpg">here</A>.
    </BODY>
    </HTML>
```

I got this message on the file I'm trying to write to.

This is my downloadFromUrl function:

```
    service.downloadFromUrl = function(url, cb) {
      var filename = uuid.v4().replace(/-/g, '');

      var stream = fs.createWriteStream('/tmp/' + filename);

      var req = request.get(url);

      req.on('end', function() {
        // check if file is valid and move to the correct folder
        cb(null, filename);
      });
      req.on('error', cb);

      req.pipe(stream);
    };
```

contra:
This is still an issue, can this be reopened?

PR #372, opened 2014-04-28
Added pipe redirection

Zulunko:
This is a fix for issue #298. Simple pipe redirection was not working; this makes it instead keep track of the stream past redirects.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -397,9 +397,14 @@ Request.prototype.write = function(data, encoding){
  */
 
 Request.prototype.pipe = function(stream, options){
+  this.stream = stream;
   this.piped = true; // HACK...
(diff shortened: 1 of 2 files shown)


contra:
@defunctzombie Interested in merging a PR for this? Pretty major issue that pipes don't work with redirects, would def send one if you OK it.

pornel:
Yes, that looks like an important fix

contra:
@pornel It looks like this was fixed by TJ already a while back. Should be reclosed.

My issue was different, attaching `'data'` listeners to a streaming request doesn't work so it isn't a real stream, it just has the pipe function. Threw me off, may need clarifying in the docs that pipe is just a sugar fn and it isn't a real stream.

PR #393, opened 2014-06-05
client: add xhr2 file download

vstirbu:
Adds ability for the client to download files as ```arraybuffer``` or ```blob```.

First register a handler that will instruct superagent how to parse the response body. 

```javascript
request.parse['myFileMimetype'] = function(obj) {
  return obj;
}
```

then, you have to set up the responseType. The xhr object is available only after the request is available, therefore the handler has to be registered for each request:

```javascript
request
.get('/arraybuffer')
.on('request', function () {
  this.xhr.responseType = 'arraybuffer'; // or blob
})
.end(function(res) {
  // now res.body is an arraybuffer or a blob
});
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -294,7 +294,9 @@ function Response(req, options) {
   options = options || {};
   this.req = req;
   this.xhr = this.req.xhr;
-  this.text = this.xhr.responseText;
+  if (this.xhr.responseType === '' || this.xhr.responseType === 'text') {
+    this.text = this.xhr.responseText;
+  }
   this.setStatusProperties(this.xhr.status);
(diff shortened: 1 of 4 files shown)


defunctzombie:
Looks good to me. What about the node side of the library? This isn't relevant for that I guess?

@vstirbu Does this PR need refreshing or is it good to go?

twoBoots:
Would be great to see this merged in!

rase-:
:+1: LGTM as well

@vstirbu looks like it needs a rebase.

vstirbu:
@defunctzombie I've made another PR on the same topic #566 .

Not the walk in the park as initially expected as the library API changed a bit. This feature is only against the client side. Node side does not have this limitation.

Issue #34, opened 2011-11-18
Synchronous vs Asynchronous


quangv:
Any way to make the request synchronous as opposed to asynchronous ?

As in I want the request to fully complete, and callback executed before it runs the next command...

```
request.get 'http://myurl/', (req, res)->
  text = res.text
console.log text
```

Also off-topic, but is req suppose to return 'null' ? it is for me. (using this on Node.)

Thanks.

PR #394, opened 2014-06-06
add option for synchronous requests to client lib

basti1302:
This would add an option to the client lib to switch the XHR to synchronous behaviour. 

I just needed this for a client that switched from some other xhr lib to the superagent client lib but had one weird feature where synchronous was definitely needed.

I do understand that synchronous ajax requests are usually a total no-go and you probably might not want to have this in superagent at all.

If you, against all odds, consider to merge this one I'd be happy to provide more changes to round this up (docs, tests).

Might be related to #34, but the discussion there is about node, I believe.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -449,6 +449,7 @@ function Request(method, url) {
   this.url = url;
   this.header = {};
   this._header = {};
+  this._asynchronous = true;
   this.on('end', function(){
     var res = new Response(self);
     if ('HEAD' == method) res.text = null;
@@ -825,6 +826,11 @@ Request.prototype.withCredentials = function(){
   return this;
 };
 
+Request.prototype.synchronous = function(){
+  this._asynchronous = false;
+  return this;
+};
+
 /**
  * Initiate request, invoking callback `fn(res)`
  * with an instanceof `Response`.
@@ -878,7 +884,7 @@ Request.prototype.end = function(fn){
   }
 
   // initiate request
-  xhr.open(this.method, this.url, true);
+  xhr.open(this.method, this.url, this._asynchronous);
 
   // CORS
   if (this._withCredentials) xhr.withCredentials = true;


SanderSpies:
Still needed when doing a request inside onunload and related events. Please re-consider adding it :-)

danieljoppi:
+1

psanders:
+1

blackbing:
+1

pornel:
This will not be added.

Synchronous XHR is an awful mistake in the DOM API. We shouldn't be getting in the way of its deprecation and eventual removal.

blackbing:
Same reason with @SanderSpies which is onunload event. While Navigator.sendBeacon can only send POST and can not set custom header is spec. I know it will be deprecate in the future, but this is an option variable in XMLHttpRequest spec. I think library should not limit the usage, developer have to know what they are doing.

pornel:
I know [sendBeacon is bad](https://github.com/whatwg/fetch/issues/124), but I still don't think it's a reason to allow sync XHR. No, no no. If you want to use a deprecated API, then use it directly.

krismeister:
Our use case is we're trying to get off MeteorJS onto a new api. Alot of meteor runs sync due to keeping the DB connection open in the node side. Our api calls need to run sync at the node level similarly.

 I know it's might be an edge case,  but it's what we're facing.

focusaurus:
AFAIK there is no synchronous networking available in regular node whatsoever. It's not possible to make a synchronous HTTP request in node. Maybe meteor is doing it via fibers but superagent only supports official node and browsers.

krismeister:
@focusaurus I didn't realize this about node, likely due to my confusion about the yield and async syntax. You're exactly right Meteor does this through fibers and their own implementation to do async promises `Promise.await(returnsPromise())`, for which Superagent worked beautifully.

PR #408, opened 2014-07-04
Fix 307 redirects

jonasfj:
Adds tests that `307` redirects with the same method.

Note, this is not really a perfect implementation. The specification says that anything other than `HEAD` or `GET` must not automatically redirect, which is what this does.

Ideally, we should add a feature similar to `-L` on `curl` which tells superagent that we trust the site and is willing follow `POST`, `PUT`, etc. redirects. Ie. it should be an option, or a user-supplied called back that is used to decide if a 307 redirect should be followed with anything other than `GET` or `HEAD`.

Note, automatically following a 307 `PUT` with a `GET` as superagent didn't before is all bad. If interested I would happily implement a `.trustRedirect(true || false)` option. Other (better) names for such a option would be great though :)

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -608,23 +608,39 @@ Request.prototype.redirect = function(res){
   // this is required for Node v0.10+
   res.resume();
 
-  // strip Content-* related fields
-  // in case of POST etc
-  var header = utils.cleanHeader(this.req._headers);
+  var headers = this.req._headers;
+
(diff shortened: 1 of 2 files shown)


defunctzombie:
@jonasfj Says I can't auto merge this anymore. Can you rebase it please. Seems to me that we want these changes.

jonasfj:
@defunctzombie rebased on current master.

PR #417, opened 2014-07-10
cope with nock not allowing connection

buildmaster:
I don't know if this is worth putting in or not but if you disable net connect using nock (which replaces standard functions and throws an error on attempting non-mocked calls) and try and do a request with a query string you should get back a NetConnectNotAllowedError but instead you get a TypeError due to req.path not existing.

This is the simplest test case I can think of but requires a devDependency on nock.js so I don't know if it'll be allowed...

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -755,9 +755,11 @@ Request.prototype.end = function(fn){
   // querystring
   try {
     var querystring = qs.stringify(this.qs);
+    if(req.path){
     req.path += querystring.length
       ? (~req.path.indexOf('?') ? '&' : '?') + querystring
       : '';
+    }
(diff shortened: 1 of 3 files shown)


samouri:
Hey all,
1. I think that the improper error bubbling is a nock issue
2. The double callback is a superagent issue.

The way most clients are probably utilizing superagent:
1. The end function is invoking the callback after trouble setting the query args: https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L642-L659 
2. The Nock overriden request emits an error: https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L531-L540

Would this be a good/generic solution? 
#1005

PR #446, opened 2014-08-29
Prefer XMLHttpRequest over ActiveX.

omeid:
There is no reason to give up XMLHttpRequest when it is available. This pull also makes it possible to use an XMLHttpRequest shim with SuperAgent.

Ref:
https://developer.mozilla.org/en-US/docs/AJAX/Getting_Started
http://www.quirksmode.org/js/xmlhttp.html
https://github.com/jquery/jquery/blob/1.x-master/src/ajax/xhr.js#L24

### DIFF
--- a/superagent.js
+++ b/superagent.js
@@ -444,7 +444,7 @@ function isHost(obj) {
  */
 
 function getXHR() {
-  if (root.XMLHttpRequest &&'file:' != root.location.protocol) {
+  if (root.XMLHttpRequest && 'file:' != root.location.protocol) {
     return new XMLHttpRequest;
   } else {
     try { return new ActiveXObject('Microsoft.XMLHTTP'); } catch(e) {}


omeid:
@tjconcept That is right. XMLHttpRequest will fail for files anyways. Only `ActiveXObject`, and that too in very limited instances support file protocol.

tjconcept:
Okay, so do we agree that with this change *less* cases are using `XMLHttpRequest`?

Namely those cases where the protocol is `file:` and `ActiveXObject` does not exist - they were previously sent to `XMLHttpRequest` but are now sent to `ActiveXObject`.

So the impact of this change is "Prefer `ActiveXObject` when the protocol is `file:` even if `XMLHttpRequest` does not exist".

omeid:
And, now it sends non-`file` requests to `XMLHttpRequest` regardless of whatever `ActiveXObject` is there or not.

tjconcept:
> And, now it sends non-file requests to XMLHttpRequest regardless of whatever ActiveXObject is there or not.

It did so before as well.

There are *only less* cases sent to `XMLHttpRequest`. No additional cases will go there, that didn't go there before. The `if` clause has been made more restrictive.

PR #508, opened 2014-12-09
getXHR made public

KidsKilla:
getXHR -> request.getXHR so you can redefine your own XHR if needed


defunctzombie:
Yea, I guess why not.

PR #520, opened 2014-12-23
Fix `Request#attach()` custom filename

xaka:


wavded:
@defunctzombie any chance we can get a release with this?

PR #524, opened 2015-01-06
Make request a promise

gossel-j:
I just added a then method to Request to make it a promise
The request follow the promise A+ rules and can be chained to other promise (return itself a promise)
Not a lot of code and lot of people who like using promises will be very glad


defunctzombie:
no. sorry.

PR #532, opened 2015-01-20
Add bower.json

phanect:
Create bower.json file.

Currently there is no `reduce` package on bower repo, but [I'm sending PR to reduce](https://github.com/RedVentures/reduce/pull/1)


phanect:
Maybe above error alerted by Travis CI is not by this change...?

PR #537, opened 2015-01-25
Small improvement to work without CORS

eyeofhell:
Hello.

Thanks for amazing lib!
I found a strange thing in CORS specification: if you add 'upload' handler, browser will trigger CORS (and send OPTIONS preflight) even for "simple" GET and POST requests (http://goo.gl/ojJq85, http://goo.gl/XB42c0). Since sometimes it's required to a
void CORS in order to minimize latency, I created a simple patch that adds 'withoutProgress()' builder method.
What do you think?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -860,6 +860,17 @@ Request.prototype.withCredentials = function(){
 };
 
 /**
+ * Disable 'progress' event that triggers CORS.
+ *
+ * @api public
+ */
+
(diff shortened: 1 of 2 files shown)


defunctzombie:
It would be better to just check if the user has added any progress listeners. If they have not, then we would not attempt to attach to upload.

defunctzombie:
Overlapping issues with https://github.com/visionmedia/superagent/pull/538. Will track there.

PR #538, opened 2015-01-27
emit 'GET' progress event

Autarc:
Trigger the progress event for GET requests in the browser.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -294,8 +294,8 @@ function Response(req, options) {
   options = options || {};
   this.req = req;
   this.xhr = this.req.xhr;
-  this.text = this.req.method !='HEAD' 
-     ? this.xhr.responseText 
+  this.text = this.req.method !='HEAD'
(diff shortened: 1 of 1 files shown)


defunctzombie:
This may have the downside of triggering a CORS preflight in some browsers.

https://bugzilla.mozilla.org/show_bug.cgi?id=727412

I think we should only connect progress handler if the user adds a progress listener in their app. Otherwise we should avoid using the upload listeners.

eyeofhell:
Thanks for update! Actually, it will trigger CORS in all browsers since this behavior is required by specs. If some browser doesn't trigger CORS it's probably something very old or not standard-compliant

defunctzombie:
Cannot merge because it will trigger CORS in browsers and users may not want that.

Autarc:
The main issue was the missing handler for emitting progress updates on GET request to resources. Although you omit the one for uploads, perhaps you could still add this one for larger downloads.

PR #539, opened 2015-01-27
Don't index array keys in querystring

travisjeffery:
Now it'll behave the same way that the querystring does by default and
should work with more parsers.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -772,7 +772,7 @@ Request.prototype.end = function(fn){
 
   // querystring
   try {
-    var querystring = qs.stringify(this.qs);
+    var querystring = qs.stringify(this.qs, { indices: false });
     req.path += querystring.length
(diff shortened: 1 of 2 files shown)


gjohnson:
+1 I think this makes the most sense. We have some breaking changes in master anyways. The fancy querystrings are too unusable by default when working with other services.

gjohnson:
@defunctzombie you cool w/ this? Gonna cut a 1.0.0-something release if you are.

defunctzombie:
@gjohnson yea, I am cool with it.

I don't remember if I merged in the first part of the error refactor (remove arity) but that would be nice to merge for 1.0.0-something.

The rest of the error stuff I have not tackled yet. I want to get close to the `success/error` behavior of jquery so that it is very clear that no error is a successful response and everything else is an error.

PR #540, opened 2015-01-30
fix:buffer content-length support

maxbbn:
fix content-length when send a buffer


defunctzombie:
Please add a test so this doesn't break in the future.

Issue #522, opened 2015-01-06
Posting JSON to PHP again


lborgman:
I use to do something like this to POST a stringified JSON:

        rq.open("POST", url, true);
        rq.setRequestHeader("Content-type", "application/x-www-form-urlencoded");
        rq.send(jsonTxt);

Then I get this as a JSON string in PHP input:

       $entityBody = file_get_contents('php://input');

When I try to do something like it with SuperAgent I get $entityBody == NULL:

       superagent.post("http://localhost:1234/myphp.php")
              .send('{"a":"b"}')
              .type("application/x-www-form-urlencoded")
              .end(function(e,s){console.log("e", e, "s", s)})

How can I send jsonTxt with SuperAgent so I can catch it with PHP as above?

lborgman:
Seems to be something with the format of the onprogress function. Switching to the supported addEventListener instead makes things work as I expected.

defunctzombie:
Closing this since it seems you resolved your own issue.

lborgman:
Thanks @defunctzombie, but that is a misunderstanding. I changed the code in Superagent to make it work. Maybe this could work in all browsers, I am not sure:

// progress
/// Fix for Chrome
if (xhr.addEventListener) {
    xhr.addEventListener("progress", function(e){
        e.percent = e.loaded / e.total * 100;
        self.emit('progress', e);
    });
} else if (xhr.upload) {
    xhr.upload.xonprogress = function(e){
        e.percent = e.loaded / e.total * 100;
        self.emit('progress', e);
    };
}

defunctzombie:
Please format the issue so we can read it. Also a PR or failing test would help explain more what is wrong here and what we should be doing. I can't tell from the code snipped what actually changed.

lborgman:
Sorry, I added the markdown to format the code.

lborgman:
I took a look again. I works with my patch (`$entityBody` in PHP above contains what I expect), but not without it (`$entityBody` is NULL).

When I look at the event it looks ok in addEventlistener, but in onprogress e.total == 0 and e.lengthComputable == false.

So it might be considered kind of a bug in Chrome, but does onprogress exist in the standards?

lborgman:
And please reopen this, of course. ;-)

lborgman:
Here is my test case, on browser side:

```javascript
    superagent.post("http://localhost:7000/test.php")
        .type("application/x-www-form-urlencoded")
        .send('{"a":"b"}').end()
```

[...]

PR #548, opened 2015-02-17
Avoid preflight OPTIONS request caused by binding xhr.upload.onprogress even when doing simple CORS requests

doesnotexist:
Fixes an issue where simple XHR requests which should not require a preflight, were still requiring a preflight due to how superagent binds xhr.upload.onprogress 

I believe this is the underlying cause of the problem in [issue 522](https://github.com/visionmedia/superagent/issues/522) 
I however ran into this issue when sending a post request to a webservice (that I do not control) which does not implement a handler for the preflight OPTIONS request. So the OPTIONS preflights always fail. The request I'm sending meets all the criteria for not needing a preflight, it is a simple POST request with no special headers and the content-type is application/x-www-form-urlencoded. Yet the way superagent binds xhr.upload.onprogress causes chrome to send a preflight. 

from the chrome source code:
"The presence of upload event listeners forces us to use preflighting because POSTing to an URL that does not permit cross origin requests should look exactly like POSTing to an URL that does not respond at all."

So the preflight can be avoided in the latest versions of chrome as long as xhr.upload.onprogress is not bound if we are doing a crossOrigin request.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -907,7 +907,7 @@ Request.prototype.end = function(fn){
   };
 
   // progress
-  if (isCrossOrigin()) {
+  if (isCrossOrigin(xhr)) {
       // do not bind xhr.upload.onprogress
   } else {
     try {
--- a/lib/client.js
+++ b/lib/client.js
@@ -61,6 +61,18 @@ function getXHR() {
 }
 
 /**
+ * Determine if XHR is cross origin request or not
+ */
+
+function isCrossOrigin(xhr) {
+  var a = document.createElement('a');
+  a.href = location.href;
+  var b = document.createElement('a');
+  b.href = xhr.url;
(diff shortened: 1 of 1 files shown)


mdawaffe:
This will prevent CORS requests from using progress events completely.  What about only adding the XHR listener if  `. hasListeners( 'progress' )` is true?

johnHackworth:
@defunctzombie is the pull request closed because you already are working on this? If not, what's the reason? I could send a fix if I know what are the prerequisites for it...

defunctzombie:
@johnHackworth I closed it because I think it is a duplicate of another open issue.

I think the PR was doing too much. Basically, we can do what @mdawaffe said and just check if there are progress listeners and only then connect progress. We should also document that this has adverse effects so users are aware.

johnHackworth:
oh, ok! makes sense... thank you for the explanation!

Issue #283, opened 2013-11-01
[discuss] change how errors are handled


defunctzombie:
Currently errors are handled in the following ways:

1. If the function to a `.get` or a `.end` as an arity of two, the first will be an error and the second a response. If it has an arity of one then only the response is provided.
1. In the case of non-200 responses, no error is emitted or set and the user must check res.status

I would like to discuss the following changes:

1. No more magic for the callback. Always require an error argument to make things simpler and more obvious at the callsites. The returned "request" object from `.get` or `.end` could still emit error to allow for some "centralize" handling.
1. All non-200 responses are treated as errors. This means that we detect the response status and make an error object with a `.status` field as needed. We do no other response processing like detecting an error message or whatever (leaving that up to the user). In the case of some network error or other xmlhttprequest error (the current error types) we don't set a `.status` since there wouldn't be one.

The above brings a few advantages in my mind (yes the changes would be major version bump breaking).

* less ambiguous function calls by removing magic arity use
* simpler error handling at the callsite since for ajax requests, any non `2xx` is a non-success and therefore an error. Making the code to act on success simpler in the common case (currently one has to constantly check `res.status` in the callbacks

defunctzombie:
This has been merged into master.

kmalakoff:
Following up https://github.com/visionmedia/superagent/pull/554 and https://github.com/visionmedia/superagent/issues/580....

The new API is non-standard for Node.js callbacks. The request was successful so there should not be an error (eg. you can check the response status code). Only if the response cannot be returned should there be an error, eg. failed to parse the body.

defunctzombie:
You could also make the claim that a non-successful http response is also
an error.

I am not married to the new api so could push out a revert it is really not
useful but there was not much pushback before and I provided lots of time.
Until something is released people don't try it.

kmalakoff:
Thank you for discussing this. I see your perspective about non-successful http responses. 

My arguments would be:

1) if there is a response then I'm not convinced this is an error but a response with a non-success status. There's potentially data with it in addition to the status code that may be handled. 

2) superagent provides a res.ok helper specifically for this purpose.

3) as a minor point around conventions/API design, I wouldn't expect a Node.js callback to return both an error and a value in the second parameter that needs to also be optionally checked. 

We would have to change 100+ calls because of this change to upgrade to 1.0.0 so let's get it right. What do other people think?

basti1302:
[...]

PR #554, opened 2015-02-23
Always provide an error argument for non-successful responses

defunctzombie:
Any completed response (non network error) also results in the error
argument being provided to the end callback. This make it easier to
handle successful responses versus failed responses. Along with the
error argument, the response argument is also supplied.

This behavior does not currently extend to `pipe()`, only to buffered
responses that are parsed in full.

see this issue for context: #283

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -298,6 +298,7 @@ function Response(req, options) {
   this.text = (this.req.method !='HEAD' && (this.xhr.responseType === '' || this.xhr.responseType === 'text'))
      ? this.xhr.responseText
      : null;
+  this.statusText = this.req.xhr.statusText;
   this.setStatusProperties(this.xhr.status);
   this.header = this.headers = parseHeader(this.xhr.getAllResponseHeaders());
   // getAllResponseHeaders sometimes falsely returns "" for CORS requests, but
@@ -468,7 +469,18 @@ function Request(method, url) {
       self.emit('response', res);
     }
 
-    self.callback(err, res);
+    if (res && res.status >= 200 && res.status < 300) {
+      return self.callback(err, res);
+    }
+
+    var msg = 'Unsuccessful HTTP response';
+    if (res) {
+      msg = res.statusText || msg;
+    }
+    var new_err = new Error(msg);
+    new_err.original = err;
+
+    self.callback(err || new_err, res);
   });
 }
 
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -767,7 +767,19 @@ Request.prototype.callback = function(err, res){
   // only emit error event if there is a listener
   // otherwise we assume the callback to `.end()` will get the error
   if (err && this.listeners('error').length > 0) this.emit('error', err);
-  fn(err, res);
+
+  if (res && res.status >= 200 && res.status < 300) {
+    return fn(err, res);
+  }
+
+  var msg = 'Unsuccessful HTTP response';
+  if (res) {
+    msg = http.STATUS_CODES[res.status] || msg;
+  }
+  var new_err = new Error(msg);
+  new_err.original = err;
+
+  fn(err || new_err, res);
 };
 
 /**
--- a/test/basic.js
+++ b/test/basic.js
@@ -51,6 +51,8 @@ describe('request', function(){
           res.error.message.should.equal('cannot GET ' + uri + '/error (500)');
         }
         assert(res.error.status === 500);
+        assert(err, 'should have an error for 500');
+        assert.equal(err.message, 'Internal Server Error');
         done();
       });
     })
@@ -83,6 +85,7 @@ describe('request', function(){
       request
       .get(uri + '/login')
       .end(function(err, res){
+        assert(!err, 'should not have an error for success responses');
         assert(200 == res.status);
         assert(2 == res.statusType);
         done();
--- a/test/client/request.js
+++ b/test/client/request.js
@@ -30,6 +30,7 @@ it('request() simple HEAD', function(next){
 
 it('request() error object', function(next) {
   request('GET', '/error').end(function(err, res) {
+    assert(err);
     assert(res.error, 'response should be an error');
     assert(res.error.message == 'cannot GET /error (500)');
     assert(res.error.status == 500);
@@ -41,6 +42,8 @@ it('request() error object', function(next) {
 
 it('request() GET 5xx', function(next){
   request('GET', '/error').end(function(err, res){
+    assert(err);
+    assert(err.message == 'Internal Server Error');
     assert(!res.ok, 'response should not be ok');
     assert(res.error, 'response should be an error');
     assert(!res.clientError, 'response should not be a client error');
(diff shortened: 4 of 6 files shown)


defunctzombie:
/cc @tj @gjohnson @rauchg @TooTallNate 

Looking for feedback on this idea and general thinking here. How are other libraries doing it. Would you like this approach? Would you prefer the current behavior and handle identifying res.status yourself.

tj:
Yea it's definitely a little annoying to handle two cases. I think the distinction between network error and transport error is important but we could hide that in the `err`. I'm pretty impartial, I think reserving `(err` to network related stuff is cleaner but usage-wise definitely annoying.

Similarly in Go the transport-level errors are just part of the response:

```
res, err := request..

if err != nil {
  // network related
}

res.Status
```

¯\_(ツ)_/¯

defunctzombie:
The current implementation (in this PR) implies the following to determine the "cause" of the error:

* `err` argument but not `res` -> network error
* `err` argument, with `res` -> user's responsibility, we provide generic error message
* no `err` argument, `res` -> successful response

defunctzombie:
My personal feelings on this are that lowlevel core libs should do what node does and what go does (errors are network things). But higher level libs can start to blur the lines about what is a "successful" response versus one that isn't. We just have to determine how high level this lib is.

For me, I would prefer code like this:

```js
.end(function(err, res) {
   if (err) {
       //.. handle errors here, or res.status == 400, etc
       return;
   }

   // good response
}
```

For some reason I find that less error prone than the current alternative:

```js
.end(function(err, res) {
    if (err) {
       return;
   }

   // oops forgot to handle res.status = 400
}
```

Alternatively, maybe this behavior could be punted to a plugin or function wrapper around the function you provide as a callback leaving the current behavior of error is network event intact and not making assumptions about what a successful response is (counter argument).

tj:
It gets weird though, because for example say we make superagent a Promise, do you really want:

```js
try {
  var res = stuff
} catch (err) {
  
}
```

when it could just be a check for `res.status`. It only really helps the callback use-case, but hinders the promise/async/await case

defunctzombie:
I don't follow the promise try/catch issue but maybe I don't know how promises behave in this respect enough.

Promises are not available in browsers natively and I am not sure I want to bloat by including some promise library. Requiring that users shim their browser also seems not so nice.

tj:
Oh yeah, forgot about browsers sucking. I guess I'm just meh in general then, whichever sounds fine to me

gjohnson:
Agree w/ @defunctzombie. At this level I think it makes the most sense and is what most users expect, maybe make it more pluggable later like you mentioned.

lib/client.js
```diff
@@ -467,7 +467,14 @@ function Request(method, url) {
       self.emit('response', res);
     }
 
-    self.callback(err, res);
+    if (res && res.status >= 200 && res.status < 300) {
+      return self.callback(err, res);
+    }
+
+    var new_err = new Error('Unsucessful HTTP response');
```
gjohnson: nitpick: maybe set error message from `http.STATUS_CODES` ( https://github.com/joyent/node/blob/master/lib/_http_server.js#L40)

tj:
If we went that route I'd almost argue that only having `(response)` makes the most sense. HTTP gives you a response that is an error or not, not an error _and_ a response, so maybe just `res.error`, but that breaks the node-style callback convention

tj:
changed my mind, I'm +1. It's hard to rationalize either since callbacks are stupid in the first place

lib/client.js
tj: +1

defunctzombie: What specific string do you have in mind? Or just the single name corresponding to the error code?

gjohnson: Yup, code -> message.

tj: something like `'<code> "<status code string>"'`might help clarify that it's an http response error, for some of the terse ones like "Conflict"

defunctzombie: Should the same thing happen on the client where this list is not available? Would require shipping this list. Or I can just change it to `Request failed`.

sadjow: I think that bringing the object `exports.STATUS_CODES` into the `superagent` library will aggregate value, since that the map object of status codes is for a stable protocol.

How can I treat the error? How can I know if that error is a transport error or an HTTP Error?

I think we could create a `new HTTPError('<code> "message"')` and maybe a `NetworkError` too. So, by example: we can retry the request if the error is a `NetworkError`.

sadjow: retry a request is a responsibility of the user.

lib/client.js
```diff
@@ -467,7 +467,14 @@ function Request(method, url) {
       self.emit('response', res);
     }
 
-    self.callback(err, res);
+    if (res && res.status >= 200 && res.status < 300) {
```
sadjow: I know that redirect status codes are being treated before this line by superagent, but status codes >= 300 until < 400 are not http errors. :)

defunctzombie: They are errors for ajax requests because they don't have bodies themselves or tell you much about the success of the request.

lib/client.js
defunctzombie: Why? You know if it is a network error or not based on if there is a `response` object in the callback.

We could add a flag to the `error` object to let you know more but my thinking was that just checking if you have a response (after error) would let you know more about the situation.

defunctzombie: I think doing this in node is a no brainer, but on the client I am less sure mostly cause of the giant string dump.

defunctzombie:
The PR has been updated to use `http.STATUS_CODES` on the server and `xhr.statusText` on the client for the error message if available. This provides the nice default error messages we want.

kmalakoff:
I don't think this is a good API change. We are totally happy checking the res.status and leaving errors for actual errors like a body parsing error.

What do other people think?

defunctzombie:
@kmalakoff please also see https://github.com/visionmedia/superagent/issues/283 as I posted in the original comments above

nw:
I think the change is a tad unintuitive and mixing layers. My mental mapping has been error is when something bad happened with the communication itself. Sure a 500 is bad but the server responded and the response is well formed. It is up to my app to handle. This for example could be throttling requests when getting a 420 from twitter, its not that my request was bad. It is a state "above" the TCP layer that needs to be resolved.

I would be more inclined to agree with this change if every server used status codes appropriately, they don't.

rauchg:
It's a really bad idea to consider HTTP-level error codes actual errors. Maybe a better way to think about it is, is a stack trace helpful at all for a 404? That's certainly not an `Error`.

kmalakoff:
@defunctzombie has chosen to stay the course on this change despite:

1) that the change breaks the Node callback conventions
2) is against the original design philosophy of superagent 

Forked: https://github.com/kmalakoff/superagent-ls and https://www.npmjs.com/package/superagent-ls. I'll aim for a first release the weekend.

I'm happy to accept for issues and pull requests for this fork. 

Please star [the superagent-ls repo]( https://github.com/kmalakoff/superagent-ls ) to show your support for reverting the 1.x API would be great, too  :smirk:

kmalakoff:
superagent-ls has been released on npm: https://www.npmjs.com/package/superagent-ls and component.io

Please star and use [the superagent-ls repo]( https://github.com/kmalakoff/superagent-ls ) to show your support for reverting the 1.x API  :smirk:

PR #555, opened 2015-02-23
Re-enable IE9 tests

rase-:
- Includes a `btoa` polyfill for the basic auth test -- where should we note that this is necessary for basic auth in IE9?
- Switches a `should` on a number to an `assert` -- failed on IE9
- Doesn't run the `PATCH` test on IE9 (feature detection) due to a weird response indicating that a Jetty server doesn't support the method -- this is likely a SauceLabs issue
- Doesn't run the xdomain failure test (feature detection) as afaik this isn't really supported atm and fails on IE9

Client tests are passing, but I'm getting a server test fail (also on master, though).

Thoughts?

@defunctzombie

### DIFF
--- a/.zuul.yml
+++ b/.zuul.yml
@@ -12,4 +12,4 @@ browsers:
   - name: android
     version: latest
   - name: ie
-    version: 10
+    version: 9..10
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "mocha": "*",
     "should": "3.1.3",
     "zuul": "~1.19.0",
-    "browserify": "~6.3.2"
+    "browserify": "~6.3.2",
+    "Base64": "0.3.0"
   },
   "browser": {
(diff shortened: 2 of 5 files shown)


rase-:
Here's the stacktrace for the failing test:
```
  1) request on redirect should follow Location even when the host changes:
     Uncaught TypeError: Cannot call method 'send' of undefined
      at null.<anonymous> (/home/rase/OSS/superagent/test/node/redirects.js:35:9)
      at timer._onTimeout (timers.js:219:16)
      at Timer.listOnTimeout [as ontimeout] (timers.js:112:15)
```

Happens on both `0.10` and `0.12`.

rase-:
Though it seems fine on travis.

rase-:
https://github.com/visionmedia/superagent/pull/556 fixes that aforementioned test for me.

Issue #178, opened 2013-02-04
Follow redirections when piping


buschtoens:
```javascript
var request = require('superagent')
  , fs = require('fs');

var stream = fs.createWriteStream('/apparently/not/the/redirected.file');
var req = request.get('/I/will/redirect');
req.pipe(stream);
```

When the requested resource performs a redirect, the response stream (`req.pipe`) is empty and nothing is written to `stream`.

rase-:
Looks like it was addressed recently: https://github.com/visionmedia/superagent/pull/558

cc @defunctzombie

thanpolas:
mmm indeed, this looks promising

PR #558, opened 2015-02-25
Fix redirect support for pipe method.

cyrilis:
Fixed redirect support for pipe method, I've stumbled on this issue for a while.  
Max redirects can be set with .redirects(n), just be the same with 'end' method.
Hope this helps.  :smile:


buschtoens:
Fixes #178.

PR #559, opened 2015-02-26
Add IE11 back to tested browsers

rase-:
Blacklisted a couple tests have some weird errors, presumably from SauceLabs as they result in Jetty error messages.

`make test-browser` all green.


defunctzombie:
LGTM

Issue #172, opened 2013-01-19
method signatures (ie. .get(...), .post(...), etc.) different in client and node version


MatthewMueller:
in the client version the signature looks like this:

    request.post(url, data, fn)

while the node version the signature looks like this:

    request.post(url, fn)

not a huge deal because you can just build these requests long-hand, but it'd be nice to update the node API to make it consistent.

PR #561, opened 2015-02-26
Make method function shorthand consistent

rase-:
Implemented the `request.post(url, data, fn)`-type shorthand for all method functions.

Unsure if this is the best place to add the new tests.

Fixes https://github.com/visionmedia/superagent/issues/172

cc @defunctzombie

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -1052,8 +1052,10 @@ function request(method, url) {
 methods.forEach(function(method){
   var name = 'delete' == method ? 'del' : method;
   method = method.toUpperCase();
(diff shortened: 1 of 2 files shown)


defunctzombie:
I merged this in for parity with the client side for now. However we should just remove this way of making requests in favor of the separate function call approach which is clearer at the callsite about what happens anyway.

rase-:
@defunctzombie completely agreed. I figured for now it's good to have parity, but using the builder API for setting things up is so much more composable and clean that we might as well eventually drop this.

PR #563, opened 2015-02-27
Redirect to a correct path for relative location paths

### LINKED ISSUE #542: Redirect with relative location fails
http://www.unizzang.com returns 302 with 'location: shop/index.php'
and
http://www.unizzang.com/shop/index.php returns 302 with 'location: main/index.php'

Final url must be 'http://www.unizzang.com/shop/main/index.php',
but superagent tries 'http://www.unizzang.comshop/index.php' and 'http://www.unizzang.commain/index.php'.

I think Request.prototype.redirect should use url.resolve to make final url.

sixmen:
Fixes #542

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -9,6 +9,7 @@ var FormData = require('form-data');
 var Response = require('./response');
 var parse = require('url').parse;
 var format = require('url').format;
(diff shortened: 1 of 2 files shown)


rase-:
LGTM

cc @defunctzombie

PR #566, opened 2015-02-28
xhr2 file download take two

vstirbu:
The API changed somehow and here is another take on #393 against the current HEAD.


defunctzombie:
@rase- thoughts?

RileyTomasek:
@vstirbu thanks for doing this! might be worth adding to the docs somewhere @defunctzombie?

Is there an easy way to do this in node? I'm writing an SDK that works in both node and the browser and I'm struggling with the node part here.

Issue #442, opened 2014-08-15
Triggering OPTIONS preflight in firefox


rralian:
We're using CORS to fetch a cross-domain JSON file with superagent via GET request, and we noticed that a simple CORS request that should not trigger a preflight OPTIONS check was still doing so in firefox. It seems to come from the progress check for uploads https://github.com/visionmedia/superagent/blob/master/lib/client.js#L908-L913

mdawaffe:
See also #548.

PR #567, opened 2015-03-03
Add xhr upload listener only if present on the request

rase-:
This way we can avoid awkward OPTIONS requests.

Should deal with https://github.com/visionmedia/superagent/issues/442

cc @defunctzombie


defunctzombie:
beautiful!

PR #569, opened 2015-03-04
add jspm support

u9520107:
add jspm dependencies and main configuration in package.json to support jspm package manager uses

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,15 @@
     "reduce": "reduce-component",
     "./test/support/server.js": "./test/support/blank.js"
   },
+  "jspm": {
+    "main": "./lib/client.js",
+    "config": "./config.js",
+    "registery": "jspm",
+    "dependencies": {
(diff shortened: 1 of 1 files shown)


defunctzombie:
Don't see the benefit in this over existing tools that read what is already in package.json.

u9520107:
@defunctzombie So if I use jspm as client-side package manager, I have no other choice than to fork the project and add those properties in my fork?

defunctzombie:
@u9520107 yes. I think that all of the relevant metadata is already in package.json and other package managers make us of that information. It would be unwise to throw every custom package manager tidbit into the package file since everyone has their favorite. As far as I am aware webpack and browserify work with the format we are using.

u9520107:
@defunctzombie Ok, thx. I suspect that jspm should support the "browser" property. Might be a bug that it's not currently working. I'll go probe around jspm's issues.

PR #572, opened 2015-03-05
Set the default response mime type to text/plain as specified by RFC1341.

warrenseine:
If no Content-Type is specified by the server, superagent does not set the MIME type and the response does not get parsed at all. Neither `response.body` nor `response.text` are set.

[RFC1341](http://www.w3.org/Protocols/rfc1341/4_Content-Type.html) (or simply [Wikipedia](http://en.wikipedia.org/wiki/MIME)) documents that the default MIME type should be `text/plain` (superagent defaults to an empty string).

Using this patch, superagent can now fill the response with the default text parser even if the server does not set the Content-Type.


warrenseine:
Ok, good catch Travis. I will try to fix this.

PR #575, opened 2015-03-09
Update mime package to support browserify

gagoman:
We can not use `superagent` with `browserify` without additional transform (`brfs`) because of `readFileSync`.

Latest version of `mime` uses `mime-db` and compiles well via `browserify`.


gagoman:
`mime` package still uses `readFileSync`, so this PR is useless.

PR #576, opened 2015-03-09
Document removal of end() arity check, err & response

danharper:


defunctzombie:
amazing, I had not gotten around to doing the docs yet!

defunctzombie:
BTW. I think you need to edit this file (https://github.com/visionmedia/superagent/blob/gh-pages/docs/index.md) otherwise your changes will be overwritten next time docs are generated.

danharper:
Ah of course! - I'll do that now.

PR #582, opened 2015-03-12
Adds progress events to downloads

jeffkole:
Would this be a reasonable solution in place of #538?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -909,12 +909,18 @@ Request.prototype.end = function(fn){
   };
 
   // progress
+  var handleProgress = function(e){
+    if (e.total > 0) {
+      e.percent = e.loaded / e.total * 100;
+    }
+    self.emit('progress', e);
+  };
+  if (this.hasListeners('progress')) {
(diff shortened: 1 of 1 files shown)


defunctzombie:
Can you describe a bit about when this would be used and what this is for?

jeffkole:
Sure.  The way we are using it is to show a progress meter as we download files (PDFs, Word docs, etc) from S3 for display in the browser.

lib/client.js
```diff
@@ -909,12 +909,18 @@ Request.prototype.end = function(fn){
   };
 
   // progress
+  var handleProgress = function(e){
+    if (e.total > 0) {
+      e.percent = e.loaded / e.total * 100;
+    }
+    self.emit('progress', e);
```
defunctzombie: What if `total < 0` ? We won't have `e.percent`. Do we want to have a default value or stick with undefined?

jeffkole: I was looking to prevent the potential divide by zero error, though the result would just be `Infinity`.  Perhaps that is sufficient safety for the API.

Issue #591, opened 2015-03-14
[discuss] allow creation of separate instances of superagent


kmalakoff:
We have been struggling with a problem with parsers being global as we use superagent as a basis for internal libraries. 

The problem is that one developer may register a parser for XML as being xml2js and another one just wants to use the XML text, but if both libraries are included, one stomps on the other depending on require ordering.

I saw in a previous issue: https://github.com/camshaft/superagent-defaults where they provide a way to create and set defaults for instances of superagent. 

I was wondering what people thought about making the instance creation a first class implementation in superagent (perhaps with reduced or different functionality than superagent-defaults) and extending it to include parsers and serializers.

Global use (as it is now)

```
request = require 'superagent'
request.parser['application/xml'] = request.parser.text # (server only). Now we could have stomped on a dependent library's 'application/xml' parser

request.get('/url.xml').end (err, res) ->
request.get('/url2.xml').end (err, res) ->
```

Instance use (new solution)

```
request = (require 'superagent')() # create instance
request.parser['application/xml'] = request.parser.text # (server only). 'application/xml' parser only used on all requests in this module

request.get('/url.xml').end (err, res) ->
request.get('/url2.xml').end (err, res) ->
```

We have encountered this problem internally and switched to doing the parsing in the end function to make the selection of the parser reliable. It think this has a potential to be a problem as more libraries use superagent.

What do other people think?

Maybe another approach would be to make deprecate globals and instead introduce async middleware for processing responses (like express) in addition to the current middleware for modifying requests.

defunctzombie:
Overall I am +1 on the idea. The global parsers are one of the places where the lines between module and global ajax like library are blurred.

I will say that I think that the instances should have some sane defaults as they do today with the option to override those defaults.

kmalakoff:
There is the middleware approach that also might work:

```
request.post('/url')
  .use(parser)
  .end (err, res) ->
```

Or even specialized functions:

```
request.post('/url')
  .parser('application/xml', xmlParser)
  .end (err, res) ->
```

In either case, by default superagent could first register the defaults and then overrides or new parsers could be added by the caller.

I think the question comes down to whether having globals for superagent makes sense in a world where it it used inside libraries who cannot control what others set.

Finally, we have created what we call a superagent factory which abstracts out the get, post, put, head, del functions (plus default signature variants) so that we can derive and create defaults from request.Request. This could be a way to allow library writers to define their own base Request class with their own defaults.

[...]

Issue #549, opened 2015-02-18
Should browser parsers be asynchronous?


kmalakoff:
I'm trying to use the xml2js modules on client and server, but because superagent.parser on the browser is synchronous and in node is synchronous, I cannot create a single parse function for xml.

Given the current control flow, there isn't an easy way for me to monkey patch parseBody on the browser for an async method  so I wonder if browser parsers be asynchronous in general and the client architecture updated to match.

I have another option to try to implement two different versions of xml parsers (sync in the browser and async in the server) and to ensure they are consistent, but this is a lot of work.

defunctzombie:
IMO the xml2js module parser method is deceiving you into thinking it is async. That method can do all the work it needs to in one go. It is not an async method and should expose an API that is also sync.

The question of changing the superagent API for parsers to support async parsing is potentially useful but I don't think it is something to get into at this time.

kmalakoff:
I used the xml2js example to give you something concrete to think about. 

There is a critical difference between the server and browser responses, one is stream based and the other is not. This means one is async and the other is synchronous. That means that because I am writing a cross-platform parser, I have this weird situation where the parser methods have the same names, but have different functional signatures and like in this case where I want to use an open source module that only supplies an asynchronous API, I have no way to make it work except to override the superagent end method.

By closing this, does this mean you are unwilling to track this API inconsistency problem for future consideration? I would prefer if superagent kept things consistent and simple by supplying a single asynchronous interface across browser and server.

defunctzombie:
I am not sure the browser affords supplying such an interface. What is your proposed API that the server can do but the browser does not? Do you want us to support pipe on the browser? In the browser we have the full response body, you can do with that what you wish without relying on superagent.

kmalakoff:
If you would like to take this discussion in that direction....

The browser can provide asynchronous interfaces and streams are coming. Today, we are using streams with pipe on the client through browserify/webpack, but I do not believe that is right for everyone so it should not be pushed onto them.

Of course, there are some differences like you mention that won't make sense because of cost/benefit (eg. pulling in a client side stream),  adding additional/unnecessary complexity, etc.

What I am saying is that superagent should aim to minimize the differences since besides its awesome API, one of superagent's main value propositions is around creating consistency across the client and server.

defunctzombie:
I will re-open, but I don't see this happening anytime soon. If you provide a working concept for what you propose, it might make things clearer.

kmalakoff:
[...]

PR #593, opened 2015-03-21
Querystring serialize bug

alekzonder:
incorrect serialization of nested objects

```js
{foo: {bar: {baz: 100}}}
```

to 

```
foo=[object Object]
```

- fixed  with qs module
- added some tests for nested objects

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -4,6 +4,7 @@
 
 var Emitter = require('emitter');
 var reduce = require('reduce');
+var qs = require('qs');
 
 /**
  * Root reference for iframes.
@@ -93,15 +94,7 @@ function isObject(obj) {
  */
 
 function serialize(obj) {
-  if (!isObject(obj)) return obj;
-  var pairs = [];
-  for (var key in obj) {
-    if (null != obj[key]) {
-      pairs.push(encodeURIComponent(key)
-        + '=' + encodeURIComponent(obj[key]));
-    }
-  }
-  return pairs.join('&');
+  return (isObject(obj)) ? qs.stringify(obj) : obj;
 }
 
 /**
@@ -119,18 +112,7 @@ function serialize(obj) {
   */
 
 function parseString(str) {
-  var obj = {};
-  var pairs = str.split('&');
-  var parts;
-  var pair;
-
-  for (var i = 0, len = pairs.length; i < len; ++i) {
-    pair = pairs[i];
-    parts = pair.split('=');
-    obj[decodeURIComponent(parts[0])] = decodeURIComponent(parts[1]);
-  }
-
-  return obj;
+  return qs.parse(str);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


defunctzombie:
How do you expect these to be serialized?

This opens the larger issue of using the `qs` module on the client which increases the client size (non trivially I think).

Not against this overall, just want to make sure that we benefit wisely from a larger library size.

alekzonder:
I check size with webpack

```
```

## without uglify
27.9 KB for version 1.1.0 without qs module
36.4 KB with qs module

## with uglify plugin

9 KB for version 1.1.0 without qs module
12.3 KB with qs module

defunctzombie:
BTW, you can get around this by using `qs` yourself and passing the result string to `.query`.

```js
superagent.get('/foo').query(qs.stringify({foo: {bar: {baz: 100}}}));
```

alekzonder:
If i just start using superagent and don't know about qs module, this is a hidden bug

```javascript
superagent.get('/test').query({foo: {bar: {baz: 100}}).end(...);

// GET /test?foo=[object Object]
// WTF!?
```

We can do recursive serialize and parse for GET-params.
But if qs module do same better and has no dependencies, why not use it?

Okay, you win :)

If you don't want use qs for browser, add example to documenatation and close this pull request

```
superagent.get('/foo').query(qs.stringify({foo: {bar: {baz: 100}}}));
```

defunctzombie:
I am not against making this better but we have trade offs.

alekzonder:
We can develop our stringify and parse in superagent, but why?

If qs module works good, covered by tests and has no large dependencies.

We can develop same functional, but without tests.

OR we can detect: if query object is nested, throw error 'use qs module for nested objects in query' :)

kmalakoff:
We have implemented a way to serialize and deserialize nested query parameters in BackboneORM with little code, but we also force a strict JSON interpretation of query parameters to allow for couch-like query syntax by URL.

serialize: https://github.com/vidigami/backbone-orm/blob/master/src/lib/json_utils.coffee#L75
deserialize: https://github.com/vidigami/backbone-orm/blob/master/src/lib/json_utils.coffee#L61

I could be wrong, but are nested queries not a standard meaning wouldn't developers need to find/implement and include custom query parsers on the backend of their choice to match the query-string implementation? eg. maybe this better left to user-land

defunctzombie:
@kmalakoff I think you are correct in that nested queries are not a standardized thing and both the parser and serializer need to match.

The node side of this library already uses `qs` because size is not an issue but the client size version doesn't obviously. I am inclined to lean the behavior as is and add documentation indicating what to do when this happens. The throw could also work but I am less inclined to do that because those tend to break other js execution.

kmalakoff:
It's tough because of the current asymmetric behaviour between browser and server. People expect consistency (https://github.com/visionmedia/superagent/issues/549).

One could argue that this should be treated at the middleware or user-land level both for node and the browser to enforce consistency. If superagent is easy to configure at a library level (https://github.com/visionmedia/superagent/issues/591), it shouldn't be a problem for someone to register query-string once on browser and once on the server for all the instances they control (or whatever their case use may be) and then match the parsing with the backend of their choice.

Alternatively, instead of throw, you could log a warning with an explanation of the inconsistency for the reason of bloat-control.

Issue #580, opened 2015-03-11
Breaking API change (1.0.0): error on status 400


kmalakoff:
I just wanted to make sure this wasn't a mistake because it is breaking all of our uses of superagent with a callback like `(err, res)` so we are locking to 0.x.

Is this change to both send an error and the response the desired behaviour: https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L771

With standard node callbacks, we are used to getting an error or the response (not both) and all of our client code has been written assuming this, e.g. errors are only for things like parsing errors, not expected HTTP response codes.

defunctzombie:
This is indeed an expected breaking API changed that was discussed in an issue as well as this PR

https://github.com/visionmedia/superagent/pull/554

The thinking, to summarize, is that any non success http response is an error. If you want to know if it is a network error or something else, you can use `err.status` or see if there is a response object as well as an error.

This is documented in the changelog and the reason for the 1.x release to indicate API change was breaking.

kmalakoff:
Wow! This is not a good API change.

Down-vote. We'll stick with 0.x.

defunctzombie:
Why do you feel like that is not a good API change?

/cc @gjohnson

kmalakoff:
I've added a comment on https://github.com/visionmedia/superagent/pull/554

kmalakoff:
@defunctzombie has chosen to stay the course on this change despite:

1) that the change breaks the Node callback conventions
2) is against the original design philosophy of superagent 

Forked: https://github.com/kmalakoff/superagent-ls and https://www.npmjs.com/package/superagent-ls. I'll aim for a first release the weekend.

I'm happy to accept for issues and pull requests for this fork. 

Please star [the superagent-ls repo]( https://github.com/kmalakoff/superagent-ls ) to show your support for reverting the 1.x API would be great, too  :smirk:

kmalakoff:
superagent-ls has been released on npm: https://www.npmjs.com/package/superagent-ls and component.io

Please star and use [the superagent-ls repo]( https://github.com/kmalakoff/superagent-ls ) to show your support for reverting the 1.x API  :smirk:

PR #594, opened 2015-03-21
Superagent 2.0.0 - superagent with less suck (revert 1.0.0 API changes in end)

kmalakoff:
Let's make the superagent community happy by doing the right thing and revert the API changes introduced in 1.0.0. 

The API changes in 1.0.0:

1) break Node callback conventions
2) are against the wisdom of the original superagent design
3) break so much code depending on superagent without a clear benefit that a majority of users of superagent agreeing with it (many do and many do not). The amount of time and effort to upgrade and test libraries and applications across the superagent userbase is enormous and non-trivial. 
4) could be implemented differently using the `success/error` pattern that would not break backwards compatibility

For people who prefer the old API and want to benefit from the latest version of superagent, I've [forked superagent to superagent-ls (for less suck)](https://github.com/kmalakoff/superagent-ls).

Fork: https://github.com/kmalakoff/superagent-ls
Released on npm: https://www.npmjs.com/package/superagent-ls and component.io.

Please star and use [the superagent-ls repo]( https://github.com/kmalakoff/superagent-ls ) to show your support for reverting the 1.x API  :smirk: 

The ship has not sailed and there's a life raft in the short term!

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,7 @@
+2.0.0 / 2015-03-21
+==================
+* reverted API change from 1.0.0. See "All non-200 responses are treated as errors now"
+
 # 1.1.0 (2015-03-13)
 
  * Fix responeType checks without xhr2 and ie9 tests (rase-)
--- a/component.json
+++ b/component.json
@@ -2,7 +2,7 @@
   "name": "superagent",
   "repo": "visionmedia/superagent",
   "description": "awesome http requests",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "keywords": [
     "http",
     "ajax",
--- a/lib/client.js
+++ b/lib/client.js
@@ -472,16 +472,7 @@ function Request(method, url) {
       return self.callback(err, res);
     }
 
-    if (res.status >= 200 && res.status < 300) {
-      return self.callback(err, res);
-    }
-
-    var new_err = new Error(res.statusText || 'Unsuccessful HTTP response');
-    new_err.original = err;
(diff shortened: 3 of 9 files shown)


defunctzombie:
Dude, seriously please stop spamming issues like this.

> Let's make the superagent community happy by doing the right thing

You are the one that is unhappy so do not assume everyone is unhappy. There are others who are happy with the API change and new users won't care, hell they may even find it makes their code simpler for the 90% case.

Maybe the change turns out to be disliked but we don't know that yet. You have spammed enough of the posts about the topic that anyone who is looking for it will find your repo and desire to have responses all the time.

kmalakoff:
@defunctzombie I am not the only one who is unhappy with the change. You have now heard from others in the issues.

Like many others, we have chosen the superagent library as our go-to solution for HTTP on client and server. You made a significant change to the API which affects us deeply and that like myself, many people do not agree with.

I [let you know](https://github.com/visionmedia/superagent/issues/580) as soon as I found out about this change, I engaged your arguments in the issues (https://github.com/visionmedia/superagent/pull/554, https://github.com/visionmedia/superagent/issues/283), I was not convinced by your arguments, and I was forced to [fork the repo](https://github.com/kmalakoff/superagent-ls). 

Obviously, this is a tough decision to reverse because it means people who upgraded to the new API need to again revert chose changes. I get that. Unfortunately, it is also the right thing to do.

Hey, the benefit of open source is that we can all help make things better. I'm trying to help.

PR #599, opened 2015-03-24
Add support for 308 redirects (similar to 307)

michaelcontento:
This PR is very similar to #408 (Fix 307 redirects). 
In fact most things were already in place and I've only ...
* ... added 308 to the list of known redirect codes
* ... added a short comment to indicate that 308 is handled
* ... copied the 307 tests and changed it simply to 308


rase-:
LGTM. Thanks!

Issue #606, opened 2015-03-28
request.get(...).send(...) not populating querystring


yourcelf:
The [documentation](https://visionmedia.github.io/superagent/#query-strings) indicates that a call to ``request.get(...).send(...)`` should invoke ``.query(...)``, so that the data given to ``send`` populates the querystring rather than a POST body.

This doesn't seem to be working; the querystrings from a request build with ``.send`` are empty.  I'll include a PR with failing test case shortly.

focusaurus:
Docs updated to no longer indicate this API. Thanks for finding this issue!

vmasto:
Greetings, this also appears on the README of the repo (under installation)

https://github.com/visionmedia/superagent#installation

focusaurus:
The only instance of `.send()` I see in the README is for a `POST` request, which defines the request body to be sent as always. This particular issue pertains specifically to GET requests.

vmasto:
Ahh, my apologies, I naively assumed it was an all-in-one like jQuery's `data` property.

PR #607, opened 2015-03-28
Test case for request.get(...).send(...) querystring

yourcelf:
Test case for issue #606; test currently fails for ``.send``, but passes for ``.query``.

### DIFF
--- a/test/client/request.js
+++ b/test/client/request.js
@@ -293,7 +293,16 @@ it('GET request .send querystring', function(next){
   .get('/echo-querystring')
   .send({foo: "bar"})
   .end(function(err, res){
-    console.log(res.text);
+    assert('{"foo":"bar"}' == res.text);
+    next();
+  });
+})
+
+it('GET request .query querystring', function(next){
+  request
+  .get('/echo-querystring')
+  .query({foo: "bar"})
+  .end(function(err, res){
     assert('{"foo":"bar"}' == res.text);
     next();
   });
--- a/test/client/request.js
+++ b/test/client/request.js
@@ -288,6 +288,17 @@ it('request .send()', function(next){
   });
 });
 
+it('GET request .send querystring', function(next){
+  request
+  .get('/echo-querystring')
+  .send({foo: "bar"})
+  .end(function(err, res){
+    console.log(res.text);
+    assert('{"foo":"bar"}' == res.text);
+    next();
+  });
+})
+
 it('request .set()', function(next){
   request
   .get('/echo-header/content-type')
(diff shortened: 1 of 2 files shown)


defunctzombie:
`.send` won't result in a querystring, so not sure what that test is trying to show.

yourcelf:
As mentioned in #606, the [documentation indicates that it does](https://visionmedia.github.io/superagent/#query-strings):

> When issuing a GET request the res.send(obj) method will invoke res.query(obj),

This test shows the described behavior.  If that's not what is desired/expected, the docs need to be changed.

defunctzombie:
That is tragic. I think being explicit is better. Did this behavior ever
work?

yourcelf:
I don't know if it ever worked. But I know I tried it, when implementing a polymorphic function that handled both getting and posting, and the docs said I could -- something like:

    function sendIt(method, data) {
        request[method]
            .send(data)
            .set("Authorization", ...)
            ... other stuff ...
           .end()        
    }

Having to branch "send" to "query" when using a different verb makes that a fair shake less elegant to implement.  But I recognize that conflating query/body could constrain you from exercising [edge cases in the HTTP spec](http://stackoverflow.com/a/983458) like sending a body along with a get request.  I'd personally be happy with a docs fix as well.

defunctzombie:
Are you having this issue with the client or browser versions or both?

yourcelf:
Uh -- client version, I guess?  I'm using superagent in unit tests under node, and browserify-packed version in the browser.  Invoked like this in both cases:

    var request = require("superagent");
    request.get...

englercj:
I too encountered this issue recently, I have to branch for gets and do the exact same code block but replacing "send" with "query". Was there ever any movement on this?

I use both the node and the browser versions of this library.

blented:
+1 on this, couldn't figure out why send() wasn't working with get().  Forcing .query() instead makes code harder for no reason.  @englercj is your version up on NPM?

englercj:
@blented No, the logic to do what I mentioned is in my application code. I didn't change this lib.

richardaday:
+1.  Ran into this same issue.  Docs and test case says it's supported.  Turns out, it's not.  I was also creating a polymorphic function like @yourcelf explained above.

dominicbarnes:
+1

I remember this working at some point, but as long as the documentation is consistent/accurate I don't care which way this goes.

focusaurus:
OK, I have updated the docs to no longer lead users to believe `request().get().send()` will populate the query string. I agree that behavior was confusing and inconsistent and crossed the border from smart/convenient into magic and surprising. So `.query()` is for query strings and `.send()` is for bodies.

PR #610, opened 2015-03-28
Fix IE9 204 status bug

rase-:
Fixes failing `.noContent` test on IE9.

Just setting `.noContent` to use the possible falsy `1223` status code is not enough with the new error checking introduced, since `1223` is interpreted as an error (since it doesn't fit the proper status code range). 

I think we should just completely set any `1223` status to `204` (implemented in this PR). Alternatively we could specifically check for `1223` when determining if the status code is an error or not, but I feel like this is a better solution.

Thoughts @defunctzombie?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -385,6 +385,11 @@ Response.prototype.parseBody = function(str){
  */
 
 Response.prototype.setStatusProperties = function(status){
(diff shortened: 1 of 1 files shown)


defunctzombie:
IE. rofl.

PR #612, opened 2015-04-02
Correct broken Readme example.

ludovicofischer:
The response handler is passed the error as the first argument,
not the response.


ludovicofischer:
Looks like the build failed because Firefox did not start in the build environment.

PR #616, opened 2015-04-04
Updates node-form-data dependency, 0.1.3->0.2.0, to fix webpack issue.

wali-s:
SuperAgent depends on node-form-data@0.1.3, which uses node-mime@1.3.4 which has issues with webpack on the server-side (Ref: https://github.com/visionmedia/superagent/pull/615).

node-form-data@0.2.0 uses another library, mime-types@2.0.3, which does not have these issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "debug": "2",
     "reduce-component": "1.0.1",
     "extend": "1.2.1",
-    "form-data": "0.1.3",
+    "form-data": "0.2.0",
     "readable-stream": "1.0.27-1"
   },
   "devDependencies": {


defunctzombie:
Needs to depend on an exact version down to the patch level.

What effect does this have on the bundle size?

wali-s:
Fixed the version issue.

If you mean my webpack bundle, in my particular case, the bundle size is unchanged. I believe because both `node-mime` and `mime-types` were already included as dependencies for other libraries.

For more context on this issue: https://github.com/felixge/node-form-data/pull/95

Issue #613, opened 2015-04-02
Range of status codes for error is problematic


aaronshaf:
To many of us 400-level errors are _not_ actual errors. Sometimes they represent a desired state. This is problematic: 

```if (res.status >= 200 && res.status < 300)``` https://github.com/visionmedia/superagent/blob/master/lib/client.js#L480

Can we please at least make the status code range for errors configurable?

defunctzombie:
I think configurable range is something that would be nice. I would entertain adding such an API.

kmalakoff:
This change would need to handled very carefully. 

If any user of superagent can re-define the range globally, they may unintentionally break libraries they use that also have expectations on range. That could lead to library writers having to query the range to program defensibly against uncertain configuration.

To properly implement this, changing the range could necessitate the ability to create instances of superagent each with their own configurations (https://github.com/visionmedia/superagent/issues/591). 

Alternatively, the range could be set on individual requests, but that could be cumbersome.

Finally, we could just go back to the 0.x API and more strictly limit superagent to the transport layer instead of mixing in logic that is best left in the application/user-land.

defunctzombie:
This would not be global but per instance. We already have Apis for per
instance so this is not hard at all.

kmalakoff:
Interesting. Maybe this should be the answer for range mapping in general. 

If you could have people who want it add it to their request instances using existing mechanisms (add it to their instances) or those of us who don't want it, to be able to ignore the setting (skip it) or those who want it differently, set a different mapping (add It with a different range), that might make everyone happy rather than making the decision for users of superagent.

Or you have the current default range mapper added to each instance and provide a mechanism to clear it or set a different rangeMap function per instance.

If instance-based configuration is the way forward, we could create our own factory function for requests that set up instance defaults like disabling range mapping. 

Would you consider rather than using an existing monkey-patching API, adding a specific API for range mapping?

defunctzombie:
@aaronshaf @kmalakoff take a look at https://github.com/visionmedia/superagent/pull/617 and share your thoughts.

@kmalakoff we could think about doing something global but I think your factory approach is better for the cases of creating custom request instances.

My general thoughts on this honestly are that you shouldn't be making agent instances all over your codebase anyway and requests (whether for APIs or whatnot) typically funnel through request methods you setup yourself scoped for the requirements. Obviously not everyone does that but the things I have written that end up doing http requests tend towards all calling a single "request" point eventually to avoid hunting down request code all over the app.

[...]

PR #617, opened 2015-04-05
Add `.acceptedStatus()` to allow specific status codes (that won't result in an error)

rase-:
A suggestion to implement the feature discussed in https://github.com/visionmedia/superagent/issues/613

With this patch a new method, `.acceptedStatus()` is added, that let's you configure the accepted status codes in two ways:

With a function:

```
request
.get('http://localhost:8888/test')
.acceptedStatus(function(status) {
    return status >= 200 && status <= 208;
})
.end(endfn);
```

With an array syntax:

```
request
.get('http://localhost:8888/test')
.acceptedStatus(['40x', '2xx', 500])
.end(endfn);
```

The above will accept any status code that starts with 40, any status code that starts with 2 and 500.

By "accepting" a status code I mean that the specified status codes will lead to a successful respons, and others will result in an error. The default range of 2xx status codes can be reconfigured with this method.

Tests seem to fail currently due to what I suspect are issues with `localtunnel`, since using `ngrok` also browser tests are passing.

Thoughts on this API as it is? cc: @defunctzombie

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -458,6 +458,8 @@ function Request(method, url) {
   this.url = url;
   this.header = {};
   this._header = {};
+  this.statusFilter = null;
+
   this.on('end', function(){
     var err = null;
     var res = null;
@@ -477,7 +479,7 @@ function Request(method, url) {
       return self.callback(err, res);
     }
 
-    if (res.status >= 200 && res.status < 300) {
+    if (self.isAcceptedStatus(res.status)) {
       return self.callback(err, res);
     }
 
@@ -596,6 +598,79 @@ Request.prototype.set = function(field, val){
 Request.prototype.unset = function(field){
   delete this._header[field.toLowerCase()];
   delete this.header[field];
+  return this;
+};
+
+/**
+ * Checks if the received status code is expected. By default only 2xx status codes are expected and don't produce an error
+ *
+ * @param {Number} status
+ * @api private
+ */
+
+Request.prototype.isAcceptedStatus = function(status) {
+  if (this.statusFilter) {
+    return this.statusFilter(status);
+  }
+
+  // By default 2xx status codes may pass
+  return status >= 200 && status < 300;
+};
+
+/**
+ * Set a filter for accepted status codes (that don't result in an error)
+ *
+ * Examples:
+ *
+ *      req.get('/')
+ *        .set('Accept', 'application/json')
+ *        .set('X-API-Key', 'foobar')
+ *        .acceptedStatus(['2xx', '4xx', 500])
+ *        .end(callback);
+ *
+ *      req.get('/')
+ *        .set({ Accept: 'application/json', 'X-API-Key': 'foobar' })
+ *        .acceptedStatus(function(status) { return status === 200; })
+ *        .end(callback);
+ *
+ * @param {Array|Function} field
+ * @return {Request} for chaining
+ * @api public
+ */
+
+Request.prototype.acceptedStatus = function(filter) {
+  if (filter instanceof Array) {
+    this.statusFilter = function(status) {
+      for (var i = 0; i < filter.length; i++) {
+        if (typeof filter[i] === 'string') {
+          var regexStr = [].map.call(filter[i].toLowerCase(), function(x) {
+            if (x === 'x') {
+              return '\\d';
+            }
+
+            return x;
+          }).join('');
+
+          var regex = new RegExp(regexStr);
+          if (regex.test(String(status))) {
+            return true;
+          }
+
+          continue;
+        }
+
+        if (filter[i] === status) {
+          return true;
+        }
+      }
+
+      return false;
+    };
+
+    return this;
+  }
+
+  this.statusFilter = filter;
   return this;
 };
 
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -133,6 +133,7 @@ function Request(method, url) {
   this.qsRaw = [];
   this._redirectList = [];
   this.on('end', this.clearTimeout.bind(this));
+  this.statusFilter = null;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


lib/client.js
```diff
@@ -600,6 +604,63 @@ Request.prototype.unset = function(field){
...
+ *        .end(callback);
+ *
+ * @param {Array|Function} field
+ * @return {Request} for chaining
+ * @api public
+ */
+
+Request.prototype.acceptedStatus = function(filter) {
+  if (filter instanceof Array) {
+    this.isAcceptedStatus = function(status) {
```
defunctzombie: We should be setting this on prototype like everything else

lib/client.js
```diff
@@ -600,6 +604,63 @@ Request.prototype.unset = function(field){
...
+        }
+      }
+
+      return false;
+    };
+
+    return this;
+  }
+
+  this.isAcceptedStatus = filter;
```
defunctzombie: I think just saving the status filter and invoking it later is better than doing the monkey patch on `this` within this function

lib/node/index.js
```diff
@@ -296,6 +299,63 @@ Request.prototype.unset = function(field){
...
+ *        .set({ Accept: 'application/json', 'X-API-Key': 'foobar' })
+ *        .acceptedStatus(function(status) { return status === 200; })
+ *        .end(callback);
+ *
+ * @param {Array|Function} field
+ * @return {Request} for chaining
+ * @api public
+ */
+
+Request.prototype.acceptedStatus = function(filter) {
```
defunctzombie: If the code for the two is the same, can we start to do some more code sharing? I know we don't do that now but this could be a nice time to re-use common code. Not a blocker tho.

rase-: For sure. It is the same yeah. Maybe in a separate PR right away after merging this in?

lib/client.js
rase-: Didn't we want to set the range per instance?

lib/client.js
rase-: The idea here was to save it to the object property, and it's actually called later in the range check when the response is received, so I'm unsure what you're referring to by saving.

kmalakoff:
Good stuff. 

We should probably try to keep the code to a minimum. Some feedback:

1) Perhaps the regex code should be an optional, user-land thing that should just be wrapped in a function that can be set using the same api. My guess is that will be rarely used.

```
statusRange = require('accepted-status-range')

req.get('/')
  .set({ Accept: 'application/json', 'X-API-Key': 'foobar' })
  .acceptedStatus(statusRange(['2xx', '4xx', 500]))
  .end(callback);
```

2) Maybe you could set the default in the constructor and make it easy to clear:

```
// By default 2xx status codes may pass
this.statusFilter = function (status) {return status >= 200 && status < 300;}

// no need to a helper method
if (!this.statusFilter || this.statusFilter(res.status)) {return self.callback(err, res);}
```

```
req.get('/')
  .set({ Accept: 'application/json', 'X-API-Key': 'foobar' })
  .acceptedStatus(-> return true) # or .acceptedStatus(null) or .acceptedStatus() all work
  .end(callback);
```

3) Not sure about whether the naming could be improved? Some ideas:

- errorStatus
- isError
- isStatusError
- statusError
- acceptStatus

None of these are great, but just some thoughts in case it spurs some better ones or increases the convictions for the current one....

defunctzombie:
I think `acceptedStatus` or `acceptStatus` are both good. `acceptStatus` is maybe better in that it is more "active" and maybe fits with existing API calls better? Think: `.send`, `.end`, `.set`.

rase-:
Can switch to `.acceptStatus`

rase-:
@kmalakoff I figured the API is a bit too primitive with the function filter alone, but I'm open to create a separate module for that. Thoughts on this @defunctzombie?

Regarding your point in 2), this is how I had coded things up before I rebased, since it seemed like a better fit conventions-wise for the file. "Unsetting" the filter seems like a good idea, though. I'd vote for changing things back. Thoughts on this @defunctzombie?

defunctzombie:
I still think we shouldn't monkey patch `this` inside of the `acceptedStatus` function call.

thanpolas:
Would you entertain a `rejectStatus()`?

Once a server returns a response, that's a response, I need to handle it.

An error would be anything happening in between me sending a request and up until I receive a response.

kmalakoff:
@rase- I wouldn't look at a simpler API as primitive. It provides a good API that matches most user's use cases with little customization boilerplate (just set a function, usually a conditional expression) and that has the necessary hooks for more advanced functionality like your regex approach. If there was a lot of boilerplate involved or specific needs couldn't be met with the simpler API, I would agree with you. 

@thanpolas there's many discussions on this topic already, but the decision has been made so we are now into workaround territory. I'm maintaining [superagent-ls](https://github.com/kmalakoff/superagent-ls)  for people who believe interpreting status codes should not be part of core, superagent is more of a transport layer, and that the node callback conventions should be adhered to.

thanpolas:
@kmalakoff :+1:

defunctzombie:
I still don't like the `this.statusFilter` patching going on.

kmalakoff:
@defunctzombie there are a multiple recommendations. Can you clarify?

Are you saying that you do not like the regex status range implementation which changes an array to a function? If so, I'm proposing that that functionality not be included in superagent since it is unlikely to be used and can be provided in a user land module.

defunctzombie:
I am not a fan of this approach:

```js
this.statusFilter = function(status) {
```

As for the regexp feature. I thought it would be nice but maybe the implementation complexity doesn't justify it? The implementation could maybe be simpler. I do think it is kinda nice to be able to specify `2xx` or something similar.

@kmalakoff do you favor any array functionality or just the function argument approach? I think some sort of array is nice to keep things simple for basic use cases.

kmalakoff:
I think superagent core should be light and provide the hooks needed to customize behavior, but not bundle seldomly used behavior that is better implemented in user land. It just adds library bulk that must be tested and maintained moving forward.

Accordingly, I believe the regex is unnecessary and unlikely to be used since it is very easy with a simple function containing a conditional to provide similar functionality.

As for the:

```
this.statusFilter = function(status) {
```

Coming from a CoffeeScript background, it just looks like setting an instance/member variable with a function and I wouldn't classify it as a (monkey) patch since it doesn't add to the external API. Maybe you would prefer naming of instance variables with underscores to indicate their internal scope and not extending the API?

```
this._statusFilter = function(status) {
```

focusaurus:
Thanks all for the effort here. I'm going to close this due to stalling without enough energy to merge over many months.

@kmalakoff if you are still motivated to try to get this functionality in, we can re-open this (or a new one given the `request-base.js` refactoring I have been doing) and discuss with the current collaborators.

Personally, I would be inclined to reject this just based on disagreeing with the basic idea that interpretation of HTTP status codes should be configurable. I think if you want "soft error" behavior for certain responses, handling them in your application code outside of superagent is OK. I believe the raw data you need from the response (headers, body) is still available so I don't think there's technical barrier to "soft errors" in your application code, but if so I'd be supportive of a PR to ensure that.

I think it's important for a developer who knows superagent to be able to predict the code behavior given the details of an HTTP response without having to scan your entire application code base to find a `.acceptedStatus` somewhere that alters behavior in a surprising way.

PR #619, opened 2015-04-06
Fix redirects to other hosts with 303, 307 and 308

jonasfj:
Also added a bunch of test cases. And removed external requests to github in `redirects-other-host.js` tests, now using localhost with different ports for different hosts.

For some reason not deleting the `host` header causes bad things to happen...
I can't actually find the place where it fails to overwrite the host header with the hostname from the url. But it has got to be there somewhere.

Anyways, existing code depends on headers being deleted. So that probably a good solution.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -619,13 +619,19 @@ Request.prototype.redirect = function(res){
   }
   // 303 is always GET
   if (res.statusCode == 303) {
+    // strip Content-* related fields
+    // in case of POST etc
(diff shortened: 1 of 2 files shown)


defunctzombie:
LGTM

jonasfj:
For the record, I've been looking through the history and it was probably me who broke this in #408.

I'm not entirely sure what is the desired behaviour for headers. Since 303 force redirects with a `GET` request, it's probably sane to clean the content related headers.
For `307` which preserves content, it's probably bad to clear the content related headers.

PR #620, opened 2015-04-07
Correct misspelt name of TJ's ferret

mikemaccana:


defunctzombie:
Critical fix! Thank you.

PR #621, opened 2015-04-09
[React Native Compat] Don't require window.location to be set

amasad:
React Native, is a browser-like environement but some browser properties
don't currently make sense in React Native. For example the location
object. However, we want our users to be able to use superagent, we even
recommend it on our website.

https://facebook.github.io/react-native/docs/network.html

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -49,7 +49,8 @@ function isHost(obj) {
 
 request.getXHR = function () {
   if (root.XMLHttpRequest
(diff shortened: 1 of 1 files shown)


vjeux:
Thanks for merging it in that quickly :)

amasad:
A new version on npm would be awesome!

brentvatne:
:+1: nice

defunctzombie:
@amasad this weekend

PR #627, opened 2015-04-15
Add request.clone()

### LINKED ISSUE #623: Cloning a request
Hello there!

I’d like to have a plugin for superagent that abstracts the OAuth2 Implicit Grant flow away from me. I’d just define some configuration (e.g. where is the authentication endpoint, what’s my client id…) and off we go. This API is what I want to achieve:

    superagent
    .get('http://some.ser.ver/whatever/resource/id')
    .oauth(myOAuthProviderObj) // see below
    .exec() // promisified end()
    .then(....) // business logic

What `.oauth()` would do:

1. Check if I have an access token somewhere
2. If I do, send the request
3. If it fails with `401`, check if I have a refresh token
4. If I do, use it to get a new access token
5. Go to 1.

Actually I’m almost finished. One problem that I have though: I might have to call the original request twice (at step 2). I can’t just call `end` again, since this test fails.

~~~
it('should send a request twice', done => {
  let req = superagent('localhost');
  req.end(err => {
    req.end(err2 => {
      done();
    });
  });
});
~~~

The other solution is to clone the original request, so that when I encounter step 2 the second time, I’ll send not the same request but a perfect copy. This is nasty though because `Object.create` doesn’t do the initialization of a new request. I ended up doing this:

~~~
function clone() {
  let doppelganger = new superagent.Request();
  doppelganger.method = this.method;
  doppelganger.url = this.url;
  doppelganger._query = this._query;
  return doppelganger;
}
~~~

Obviously this is both incomplete (what about form data or the request body when POSTing?) and implementation-specific. What I’d rather have is something along the lines of `request.clone()`.

What do you think? Do you happen to have any other solutions maybe? It’s late already :)

prayerslayer:
Fixes #623, although I’m pretty sure I missed something.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -498,6 +498,23 @@ function Request(method, url) {
 Emitter(Request.prototype);
 
 /**
+ * Allow for cloning.
+ */
+
+Request.prototype.clone = function() {
+  // make new object
+  var clone = new Request(this.method, this.url);
+  // set existing querystrings
+  clone._query = this._query;
+  // set existing headers
+  clone.header = this.header;
+  clone._header = this._header;
+  // set data
+  clone._data = this._data;
+  return clone;
+};
+
+/**
  * Allow for extension
  */
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -502,14 +502,13 @@ Emitter(Request.prototype);
  */
 
 Request.prototype.clone = function() {
-  // make new object
   var clone = new Request(this.method, this.url);
-  // set existing querystrings
+  // querystrings
   clone._query = this._query;
-  // set existing headers
+  // headers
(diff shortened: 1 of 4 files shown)


prayerslayer:
How do you debug superagent when a test fails on Android 5 on Linux and IE whatever on Windows 2012?

defunctzombie:
Not sure this is all that useful. Seems like a separate function can easily be created with the plugin system (based on your original use case description).

dwieeb:
Wishing this small, useful PR was in superagent right now.

If the modifier functions (`.query()`, `.set()`, etc) don't return new immutable request objects and `.clone()` doesn't exist, then there's no built-in way to construct a request from an existing request.

PR #634, opened 2015-04-25
enable secure protocol

muddydixon:
I want to request "TLS1" server.
So add `Request.prototype.secureProtocol` method.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -681,6 +681,19 @@ Request.prototype.ca = function(cert){
 };
 
 /**
+ * Set the secure protocol.
+ *
+ * @param {String} secureProtocol
+ * @return {Request} for chaining
+ * @api public
+ */
+
+Request.prototype.secureProtocol = function(secureProtocol){
+  this._secureProtocol = secureProtocol;
+  return this;
+};
+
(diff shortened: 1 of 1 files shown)


muddydixon:
Please merge me!

defunctzombie:
what does this do?

muddydixon:
by this p-r, we can control ssl/tsl version to request.
(now I want to connect a server only accepting tls1 protocol.)

defunctzombie:
What about the client version?

muddydixon:
i forked from v1.2.0 (exactly 94afa3ae189267a8d352d72b20edb9900148959e).

defunctzombie:
Not sure what that last comment has to do with this. I was asking about the browser side version of the library.

muddydixon:
Sorry, I use superagent in node (ServerSide).

defunctzombie:
While that may be the case, we still need to keep parity between the two versions if we can. Is this feature relevant for the client version?

muddydixon:
I don't know it, but I will check it.

kornelski:
This is no longer compatible, so I'm going to close this.

Issue #636, opened 2015-04-28
react-native: res.body is null, only res.text is working


jonaswindey:
Using react-native 0.4.1, both my request.get and request.post methods are working, but the res.body is undefined, forcing me to manually JSON.parse the res.text:

```
request
      .post('http://10.0.0.3:9091/auth')
      .type('json')
      .send({
        email: 'xxx',
        password: 'xxx',
      })
      .end((error, res) => {
        if (res.status === 200) {
          console.log(res.body); // undefined
          var token = JSON.parse(res.text).token; // working
...
```

When calling this from regular React, it's working perfectly.

rpastorelle:
Same for me. The `res.header['content-length']` is `null` & `res.type` is an empty string.

slooker:
I'm having the same error as well.

rpastorelle:
@slooker @jonaswindey You can make the change in the PR above & see if that fixes for you as well. It was the cause of my problems.

slooker:
This solved the problem for me.  Nice work. :)

isair:
I'm having this issue as well. When I do `console.log(res)` I see that content-type header is null, but when I do the request using a browser and log the response I can see that Content-Type header exists and is application/json.

defunctzombie:
I think these are issues with react-native.

jsergiu:
I'm still getting this error on firefox with latest version of react and superagent. After long debugging my work around is to add request.set('Accept', 'application/json') or whatever type you expect. This way the browser will treat the response as json even the response content type is something different.

icefoggy:
I'm still getting this on react native android. It works fine for iOS, but the res.body is null and res.type is an empty string.

niftylettuce:
Same issue here

niftylettuce:
Did anyone get this working?  I've tried doing `request.set('Accept', 'application/json')` using `superagent-defaults` for example and it still doesn't work.  iOS works fine.

niftylettuce:
@icefoggy Can you share insight if you got it fixed?

niftylettuce:
Has anyone figured this out yet?

mienaikoe:
I'm having the same issue now.
Using superagent version 1.5.0

res.body is null.
res.text is full.
the response is text/json, and I'm using a plain get request.

niftylettuce:
Use my package in NPM called Frisbee

pornel:
[...]

PR #638, opened 2015-05-03
Failed to set `Content-Type` internally with the xhr's `getResponseHeader`

rpastorelle:
In response to [an issue](https://github.com/visionmedia/superagent/issues/636) in which `res.body` is `null` ...`superagent` Failed to correctly set `Content-Type` header from the xhr with `getResponseHeader` due to case-sensitivity

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -305,7 +305,7 @@ function Response(req, options) {
   // getAllResponseHeaders sometimes falsely returns "" for CORS requests, but
   // getResponseHeader still works. so we get content-type even if getting
   // other headers fails.
-  this.header['content-type'] = this.xhr.getResponseHeader('content-type');
+  this.header['content-type'] = this.xhr.getResponseHeader('content-type') || this.xhr.getResponseHeader('Content-Type');
   this.setHeaderProperties(this.header);
   this.body = this.req.method != 'HEAD'
     ? this.parseBody(this.text ? this.text : this.xhr.response)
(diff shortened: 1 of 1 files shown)


slooker:
This solved the problem for me.

defunctzombie:
sounds like react-native is at fault here? headers are not case sensitive as far as I am aware

slooker:
I'm not sure if it's react-native's fault or not, but according to https://tools.ietf.org/html/rfc7230#appendix-A.2 section 3.2, headers definitely should be case-insensitive.

dhrrgn:
@slooker @defunctzombie It is true that `getResponseHeader` should be case-insensitive.  However, this is probably an issue with JavascriptCore (which is the engine that ReactNative uses), thus cannot be fixed by the Facebook team (without an upstream fix).  I think the best solution here is just to use the correct case.

defunctzombie:
IMO the best solution is for the engine to the fixed. This has not been a
problem in any other browser to this point. If it was not in the spec I
might feel differently but given that all the other browsers behave
correctly I am hesitant to fix this since it seems like a bug in
JavascriptCore (not exactly sure what that is by sounds relatively new and
likely to fix this bug)

defunctzombie:
I would at least file an issue there and see how receptive they are. If for
some reason they are not receptive then we can revisit.

slooker:
I agree, we should absolutely post it as an issue there.  Getting it fixed where it's actually broken is a better option than applying a bandaid further down.

rpastorelle:
I don't see the harm in using the typical case for this header. It adds no extra code and still fixes the issue.

rpastorelle:
I found & fixed the root cause in React Native — https://github.com/facebook/react-native/pull/1138

defunctzombie:
:+1:

Issue #208, opened 2013-04-14
Expose Set-Cookie response updates on redirect


damianb:
real world scenario, when changing session identifiers (good practice upon login, to prevent session fixation), Set-Cookie headers should be obeyed.

defunctzombie:
PR with tests would go a long way here. We just have limited time to get to all the issues. :)

Issue #207, opened 2013-04-14
Superagent drops cookie header on redirect, botching chained requests


katanacrimson:
Really fucking helpful feature, _thanks_.  

Works **great** for when you're redirected after authorization and need to access a page authenticated.  Great way to spend four hours trying to troubleshoot code, only to find that the http request library is pulling another dumbass move, dumping the cookies yet again.

PR #645, opened 2015-05-08
Retain cookies on redirect when hosts match

tomconroy:
The problem case for this is cookie-based authentication. In my case the Wordpress REST API `/users/me` redirects to `/users/{id}` based on authentication. This works fine in the browser (it sends the cookies to the redirected path), but on the server superagent strips the cookies and the call to `/users/{id}` results in a 401 Unauthorized response.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -603,11 +603,13 @@ Request.prototype.redirect = function(res){
 
   var headers = this.req._headers;
 
+  var shouldStripCookie = parse(url).host !== parse(this.url).host;
+
   // implementation of 302 following defacto standard
   if (res.statusCode == 301 || res.statusCode == 302){
     // strip Content-* related fields
     // in case of POST etc
-    headers = utils.cleanHeader(this.req._headers);
+    headers = utils.cleanHeader(this.req._headers, shouldStripCookie);
 
     // force GET
     this.method = 'HEAD' == this.method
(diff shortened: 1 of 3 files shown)


tomconroy:
Related to #208 #207

DullReferenceException:
Any reason why this has not been merged? The cookie behavior really needs to be fixed.

facultymatt:
+1 would love to see this merged and released. 

Testing sessions with redirects is fundamental to many websites, making this bug a major blocker in my opinion.

nchase:
Hi folks, I had this problem too. It took me over a day to figure out, lots of troubleshooting code that wasn't broken. Until the behavior is changed (if it is ever changed), it would be great to call this behavior out loudly so that other people can find the issue easily.

Looks like [node-fetch](/bitinn/node-fetch) has this same behavior, I might open an issue there too.

sayanee:
:+1:

chetverikov:
+1

chetverikov:
+1

chetverikov:
+1

lib/node/index.js
```diff
@@ -603,11 +603,13 @@ Request.prototype.redirect = function(res){
 
   var headers = this.req._headers;
 
+  var shouldStripCookie = parse(url).host !== parse(this.url).host;
```
focusaurus: Does this need to be made case insensitive to be correct?

focusaurus:
@pornel This looks OK to me and several users are keen to get it merged. OK to merge?

dominiksteiner:
+1

Issue #484, opened 2014-10-27
Cross domain error falsely thrown


rickharrison:
I believe that I have been getting some falsely thrown CORs errors.

https://github.com/visionmedia/superagent/blob/master/lib/client.js#L877

I believe that when the page is refreshed or navigated away from while an XHR request is in flight, the browser will set the status to 0. Is there a better way to detect if it is specifically a cross domain error?

palamccc:
+1, One of my xhr request takes 5 seconds to complete, and if page is refreshed before completion, i am getting cross domain violation error. All of my requests are sent to same origin domain only. From stack trace, it looks like the error is generated at https://github.com/visionmedia/superagent/blob/0.21.0/superagent.js#L1280 (Win7, Chrome 40).

It looks like this exception is generated, when page load is aborted by browser. You can see this exception in console only if you enable preserve log in developer console.

defunctzombie:
I am open to suggestions on how to detect this better.

defunctzombie:
Related reading:
http://stackoverflow.com/questions/19325314/how-to-detect-cross-origin-cors-error-vs-other-types-of-errors-for-xmlhttpreq

palamccc:
From what i understand,
When a request is aborted, it may be user interruption (user leaving the page) or due to cross domain violation error. since there is no foolproof way to detect the reason, I think its better to throw generic "Page Aborted" exception. (Developer can anyway find out about cross domain violation from web console.)

As of now code throws self.crossDomainError() for all abort cases.  https://github.com/visionmedia/superagent/blob/0.21.0/superagent.js#L1280

zackify:
this is still happeneing to me. The request goes through, then this error gets thrown. Really weird.

grrowl:
+1 if this is the case it really should be documented. We're capturing many `error.crossDomain = true` errors in Sentry but we've been diagnosing them as if we have a server issue; in fact, it might be regular run-of-the-mill connection aborted during navigation.

As it stands, I don't think we're able to tell at all (and neither is superagent?)

ksesong:
I have the same problem, and cannot find a way around it.
Has anyone found out a way to avoid interrupting the request?

kahnjw:
I am running into this issue as well.

>I am open to suggestions on how to detect this better.

[...]

PR #653, opened 2015-05-16
Generalize xhr status 0 error

ahoym:
Fixes: https://github.com/visionmedia/superagent/issues/484

**Changes:**
- Rename `.crossDomainError()` ---> `outBoundError()`
- Reword the error to blanket other causes of `xhr status 0`
- Add `info` property onto `err` that outlines common causes.
- Update tests to reflect the above.

`xhr.status` of 0 is not always a CORS error. According to the [wc3 spec](http://www.w3.org/TR/XMLHttpRequest/#the-status-attribute), other possible causes include network errors or fetch termination (page reload/redirect), the latter of which seems to be affecting many people.

I think a general blanket error for all scenarios of the xhr status 0 is that something went wrong with the outbound request and have named it as such. However I'm not completely enamored by it, so discussion and other suggestions are welcome!

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -839,14 +839,15 @@ Request.prototype.callback = function(err, res){
 };
 
 /**
- * Invoke callback with x-domain error.
+ * Invoke callback with xhr status 0 error.
  *
  * @api private
  */
 
-Request.prototype.crossDomainError = function(){
-  var err = new Error('Origin is not allowed by Access-Control-Allow-Origin');
-  err.crossDomain = true;
+Request.prototype.outBoundError = function(){
+  var err = new Error('Error with outbound request');
+  err.outBound = true;
+  err.info = 'Possible causes: CORS error, network error, fetch termination';
   this.callback(err);
 };
 
@@ -910,7 +911,7 @@ Request.prototype.end = function(fn){
     if (0 == status) {
       if (self.timedout) return self.timeoutError();
       if (self.aborted) return;
-      return self.crossDomainError();
+      return self.outBoundError();
     }
     self.emit('end');
   };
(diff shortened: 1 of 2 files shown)


AhoyM:
To avoid string matching, we can probably also attach the xhr as a property of `err`, or simply throw the `xhr` itself. This way, the user can catch and handle the `xhr status 0` in the way they see fit

defunctzombie:
This would be a breaking API change but also likely to lead to less confusion. outBoundError doesn't sound like the best terminology for this tho. In general I like solving this confusing part of the code better.

scsper:
@defunctzombie I think https://github.com/visionmedia/superagent/issues/484 says that we should call it a TransportError, not an outBoundError, and that we should bump the major version.  If we made those two changes, would we be able to merge?

defunctzombie:
@scsper sounds good to me. I think your plan in having a better message for 1.x and a different error type for 2.x would work. Run it by others via a PR when you are ready.

scsper:
@defunctzombie https://github.com/visionmedia/superagent/pull/802 contains the change for a better message.  

@pornel Are you planning on making the 2.0 change as well?  If not, I can do it.

pornel:
@scsper is there a consensus what the error message should say?

focusaurus:
Trying to push this along. I think we need:

- [ ] update PR to name this `TransportError`
- [ ] finalize the error message

I'm OK with the existing error message "Request has been terminated\nPossible causes: the network is offline, Origin is not allowed by Access-Control-Allow-Origin, the page is being unloaded, etc."

If I had to tweak it, I'd try to order the example reasons by likelihood (no data to support this beyond intuition)

"Request has been terminated and may not have been sent at all.\nPossible causes: the page is being unloaded due to navigation or reload, the network is offline, Origin is not allowed by Access-Control-Allow-Origin (CORS), etc."

pornel:
The reworded message has been in superagent for a while now. I haven't renamed  `crossDomainError`, but since the message is generic I think it's not causing confusion any more.

Issue #654, opened 2015-05-18
Support setting xhr responseType on the request


mhahn:
I want to set responseType on the xhr request, looking at the code I don't see an easy way to do this right now.

Similar to withCredentials, I'm thinking of adding a responseType which will set the responseType on the XHR request.

focusaurus:
This shipped in 1.8 and I'm adding it to the docs now.

PR #655, opened 2015-05-18
Support setting the XHR responseType

mhahn:
Proposed solution to #654

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -936,6 +936,17 @@ Request.prototype.withCredentials = function(){
 };
 
 /**
+ * Enable setting the XHR responseType.
+ *
+ * @api public
+ */
+
+Request.prototype.responseType = function(type){
+    this._xhrResponseType = type;
+    return this;
+};
+
+/**
  * Initiate request, invoking callback `fn(res)`
  * with an instanceof `Response`.
  *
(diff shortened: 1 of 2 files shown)


rylan:
+1, having this feature would be super handy

deltaidea:
:+1:

ponelat:
+1

lib/client.js
```diff
@@ -880,6 +880,17 @@ Request.prototype.withCredentials = function(){
 };
 
 /**
+ * Enable setting the XHR responseType.
+ *
+ * @api public
+ */
+
+Request.prototype.responseType = function(type){
+    this._xhrResponseType = type;
```
rjonna: You will have to use xhr.response instead of xhr.responseText at line 300.

jbq:
Works great, thanks!  I had to manually resolve conflicts however, this pull request has to be updated.  See https://github.com/jbq/superagent/commit/7a8ddf4ac193573b608e592ab3a6a436eb144f0d

brianswisher:
:+1:

lib/client.js
mhahn: to reference the response you should do: response.xhr.response

mhahn:
@pornel just rebased

pornel:
The tests are failing: Error: InvalidStateError: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable. (https://jmxihwatjo.localtunnel.me/__zuul/test-bundle.js:509)

mhahn:
hm, they're passing locally for me

mhahn:
is there a way i can repro that locally? i can't even see which test failed

pornel:
It looks like this code is not compatible with internet explorer 10 on Windows 2012:

https://travis-ci.org/visionmedia/superagent/jobs/96181019

it passes in other browsers.

focusaurus:
Not sure the exact history of things here, but a `responseType(name)` method has made it into superagent releases. OK to close this out as resolved/duplicate?

PR #681, opened 2015-06-19
add tls options

cgx9:
i need use key and cert at self-sign certificate for https

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -720,7 +720,7 @@ Request.prototype.request = function(){
   options.ca = this._ca;
   options.agent = this._agent;
   for(var item in this._tlsoptions){
-    options.item = this._tlsoptions[item];
+    options[item] = this._tlsoptions[item];
   }
 
   // initiate request
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -679,6 +679,10 @@ Request.prototype.ca = function(cert){
   this._ca = cert;
   return this;
 };
+Request.prototype.tls = function(options){
+  this._tlsoptions = options;
+  return this;
+};
 
 /**
  * Allow for extension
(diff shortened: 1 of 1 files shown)


pornel:
It allows arbitrary options, and tls options can override all other properties, e.g. `tls({host:'xxx'})`, which doesn't seem right. I'd prefer an API that is limited to *only* specific useful whitelisted options.

timon:
This PR works for me, 

and I also find it concerning that current implementation will happily override any options.

At least `cert` and `key` should be exposed (I believe they could be made separate arguments), maybe some tis peer checking callbacks could be exported as well.

This is sample naïve code I have used (that's basis upon which I'm building up now):
``` javascript
var fs = require('fs');

var url = 'https://server.local/client_auth';

var myCA = fs.readFileSync('server-ca.crt');
var myCert = fs.readFileSync('client.crt');
var myKey = fs.readFileSync('client.key');

request.get(url).
  ca(myCA). // <-- already present
  tls({ key: myKey, cert: myCert }). // <-- required to  work
  end(function(err, res) {
  if (err) {
    console.log("Err: ", err);
    return;
  }
  console.log(res);
});
```

Whoaa512:
This approach seems most amicable and forward thinking in the case that the options for `tls.connect()` change.

I agree that overriding the other options is troublesome, however this could easily be avoided by simply moving the loop above the property assignments which should not be overridden.

The only other suggestion would be to add a comment above the method for documentation purposes and maybe a test. 

@cgx9, you could adapt my test from https://github.com/visionmedia/superagent/pull/832

ericnewton76:
ca(...) and tls(...) shouldn't be on the returned object from get(...) but should be on request(...) so that wasteful redundancy in testing code is reduced

matteofigus:
Hi, any news about this? It would be a good feature to add for me. Thanks for the great work.

pornel:
This will not be merged in the current state, because it allows change of arbitrary options on the agent, and we don't want to support *everything*, just properties we know are useful *and have tests for*.

- Please make it only accept whitelist of properties (without rejectUnauthorized #832)

- Please add tests for the properties accepted

pornel:
Closed in favor of #1057

If you'd like to add more options, see https://github.com/visionmedia/superagent/pull/681#issuecomment-209345788

niftylettuce:
v5.1.1 released that resolves this issue

PR #685, opened 2015-06-24
Support Certificate, Private key and CA certificates in PFX or PKCS12…

EugeneKorshenko:
Support Certificate, Private key and CA certificates in PFX or PKCS12 format to use in https request


Harishgolla:
Any idea why this PR is closed?
 We want to send pfx and passphrase  to superagent. Let us know how we could do this in superagent.

PR #689, opened 2015-07-01
Support JSON-API

schtibe:


defunctzombie:
why? is this a commonly used standard?

topaxi:
Yes, yes it is: http://jsonapi.org/

defunctzombie:
Website doesn't make it a commonly used standard, just makes it a website :) Adding media types is easy since superagent exposes this for the user. I think I would like more how the request module does it by letting you specify `json: true` option to force decoding of the response as json. Maybe adding a `json(true/false)` or similar API to force/unforce automatic decoding. Otherwise we are in a constant battle to "intelligently" detect all the various mime types for json.

topaxi:
Sure, but there seems to be a movement in a lot of web frameworks in using this jsonapi.

Ember.js is going to use the jsonapi as default: http://emberjs.com/blog/2015/06/18/ember-data-1-13-released.html
The Django REST Framework has an implementation: http://drf-json-api.readthedocs.org/en/latest/
Ruby on Rails is going to use it too, as rails-api is going to follow the jsonapi by default: http://wyeworks.com/blog/2015/4/20/rails-api-is-going-to-be-included-in-rails-5/

It would be nice if superagent, especially because it is used in browsers too, could support this, imho very small, feature.

defunctzombie:
seems mostly rails/ember specific for now. I am not against having a way to force json parsing, but as I said I think the better approach is to hint to the instance that the response should be handled as json.

PR #693, opened 2015-07-04
support for use inside webworker

lucidNTR:
there is no window object inside webworkers


tnrich:
Hey there, is anything stopping these changes being merged in? I need them to use superagent in a webworker.

tadjik1:
Yep, merge it please.

defunctzombie:
Why is this different from: https://github.com/visionmedia/superagent/pull/673

And neither have tests.

lucidNTR:
both are the same, i did not see the other pull request due to its generic description.
the other version is probably better because using self can be better handled by linters.

PR #695, opened 2015-07-06
Serialize JSON variations content types

nilgradisnik:
HTTP header Content-Type can contain variations of `application/json` media type.

See the [IANA list](http://www.iana.org/assignments/media-types/media-types.xhtml) for more information. This enables parsing of popular additional media types such as [JSON Patch](http://jsonpatch.com/) and [JSON API](http://jsonapi.org/).

The regex used that matched this schema is `/(application\/(.*)\+json)/`,

Fixes #660 and #689.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -171,6 +171,22 @@ request.types = {
  };
 
  /**
+ * Content subtype serialization.
+ * http://www.iana.org/assignments/media-types/media-types.xhtml
+ *
+ *     superagent.serializeSubtypes('application/json-patch+json') = function(obj){
+ *       return 'generated json here';
+ *     };
+ *
+ */
+
+request.serializeSubtypes = function(type) {
+  if (!type || typeof type !== 'string') return null;
+  if (type.match(/(application\/(.*)\+json)/)) return JSON.stringify;
+  return null;
+};
+
+ /**
   * Default parsers.
   *
   *     superagent.parse['application/xml'] = function(str){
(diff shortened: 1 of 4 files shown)


defunctzombie:
please follow existing coding style

nilgradisnik:
I assume you were referring to the `if` and `else if` style. I fixed that.
Also I'm not sure why Travis CI tests are failing, all the tests pass on my machine.

alinz:
@nilgradisnik I'm having a same problem with my PR! All the tests pass but Travis fails.

defunctzombie:
Travis will fail because of saucelab keys being encrypted for the repo.

defunctzombie:
please squash the commits :)

nilgradisnik:
Done.

defunctzombie:
Looking at this pr a bit more. I like that it handles the charset part of content-type but I don't like that it hard codes the regex check. There is already an object for the different serializers. Why not just add the content-type you want to serialize to that? Alternatively, if there is a convention/standard around having mime-type followed by `+more stuff` I would be open to separating the core mime type from the other stuff and checking serializers for that. This seems more general purpose.

nilgradisnik:
I agree with you. I was struggling to somehow incorporate this regex to the `serialize` map because it's a map, so we would need to add all the possible type variations in there which seems bad, where this regex solved this problem more elegantly.

Perhaps changing that `serialize` to a function that takes the content-type value and returns a serializer function seems like a better solution? This would require a bigger change that's why I decided not to do it.

I am all for having a more generic solution, if you look at the [IANA list](http://www.iana.org/assignments/media-types/media-types.xhtml) there are a bunch of `+xml` types which can be parsed by an XML parser.

This pull request was an effort to solve the JSON type variations but I can have another go at this and come up with a generic solution.

nilgradisnik:
@defunctzombie I took another pass at this pull request. I added the `serializeSubtypes` function that takes care of this. This is where we can handle serialization of any other types such as XML etc.

pornel:
I think we've added that a while go.

PR #697, opened 2015-07-07
fixed #696

### LINKED ISSUE #696: this.xhr.getResponseHeader('content-type') returns empty with react-native >0.5
I have been struggling with why all my apis start to fail after upgrading to `react-native` 0.5 and above. What I found is that, after upgrading `getResponseHeader('content-type')` returns null.

refer #181

alinz:
added some code to check `getResponseHeader('content-type'…)` prior to set to `content-type`

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -305,7 +305,10 @@ function Response(req, options) {
   // getAllResponseHeaders sometimes falsely returns "" for CORS requests, but
   // getResponseHeader still works. so we get content-type even if getting
   // other headers fails.
-  this.header['content-type'] = this.xhr.getResponseHeader('content-type');
+  var tryAnotherContentType = this.xhr.getResponseHeader('content-type');
+  if (tryAnotherContentType && tryAnotherContentType.length > 0) {
+    this.header['content-type'] = tryAnotherContentType;
+  }
   this.setHeaderProperties(this.header);
   this.body = this.req.method != 'HEAD'
     ? this.parseBody(this.text ? this.text : this.xhr.response)


defunctzombie:
don't understand the issue here. If the content-type header is null then the null value will be set to `this.headers[content-type]`. Alternatively you just don't set the value. Both seem like essentially the same behavior, no?

alinz:
hmm, what I found was, [parseHeader](https://github.com/visionmedia/superagent/blob/master/lib/client.js#L304) parse the header correctly which contains `application/json` but we are blindly trusted [this.xhr.getResponseHeader('content-type');](https://github.com/visionmedia/superagent/blob/master/lib/client.js#L308) and override the `content-type`. So in my PR, I check the return value of `this.xhr.getResponseHeader('content-type')` first, before overriding the `content-type`.

defunctzombie:
#696 sounds like a react-native bug to me. Why does it not return the correct value for getResponseHeader? Seems like if it worked as expected for that method then there would be no issue.

alinz:
Agreed, however, do you think it's a good idea to overriding header's `content-type` without checking the `xhr.getResponseHeader`'s value?

defunctzombie:
Kinda since one would reasonable expect that function to return the header value or null based on the docs: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#getResponseHeader()

focusaurus:
I'm going to close this as #696 has been closed and the maintainers have not been compelled to merge after almost 8 months. If this is still an issue for you with the latest react-native, maybe follow up with that project. Thanks for your effort, though!

Issue #670, opened 2015-06-02
arrays in query parameter are handled differently in node vs browser libraries


davis:
`.query({ hello: ['a', 'b', 'c'] })`

returns different things
node:
`?hello=a&hello=b&hello=c` (correct)

browser:
`?hello=a,b,c` (incorrect)

davis:
can be merged with #629

aalpern:
:+1: 

I just ran into this problem as well, in a REST API client I wrote that's intended for both node.js and browser usage. It means avoiding superagent's built-in query string encoding altogether, which really defeats the purpose of having it at all.

defunctzombie:
Can you make a PR if there isn't one already? Also please add tests.

aalpern:
I'll see what I can do. Although I'm currently getting 35 test failures under node.js (in node 0.12.2 and 0.12.5) from a fresh clone of `master`, so that doesn't give me a lot of confidence in testing my own changes.

aalpern:
Digging in, it looks like the bug is actually in the `qs` module, which handles all query string formatting: https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L823 , which has had 6 releases since the one in user here. 

Ah, there's an array format option in qs 2.4.0 that does the trick (and there's already a test for this case in the node.js tests, just not one in the browser). PR incoming shortly.

aalpern:
Just submitted a PR - if it could be package in a 1.2.1 release that would be :sparkles: :guitar: :tada:

ianks:
Why was this syntax (`?hello=a&hello=b&hello=c)` decided on over:

`?hello[]=a&hello[]=b&hello[]=c`

It seems to me that the latter is the most common syntax for representing arrays in a query string.

pornel:
1. If you need to, you can achieve a `hello[]` keys by using exactly that name in the input, but it wouldn't work the other way — if superagent always added `[]` then it wouldn't be compatible with servers that require `[]`-less arguments.

2. Node already does it this way, and superagent aims to be identical in browser and node.

PR #698, opened 2015-07-07
Use qs for query string serialization in the browser (addresses #670)

aalpern:
This addresses issue #670 by using `qs` to serialize values in the browser just like the node implementation does, rather than a hand-rolled loop of `encodeURIComponent`. 

This changes one test case - `{foo: null}` no longer serializes to `''`, it serializes to `foo=`, which matches the node version (and is generally what would be expected).

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -4,6 +4,7 @@
 
 var Emitter = require('emitter');
 var reduce = require('reduce');
+var qs = require('qs');
 
 /**
  * Root reference for iframes.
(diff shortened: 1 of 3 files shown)


defunctzombie:
The reason I have been hesitant to do this is because `qs` is a large module and now everyone has to bear the cost of increased file size. I know this is an annoyance but also have to balance with file size :/

aalpern:
Is it possible to structure a plugin to replace the built-in serialization? That way people who need it can opt for it, and those who don't need any more than basic key/value serialization can use the default. 

For anyone who actually needs node/browser compatibility though, they're going to have to pay the cost anyway (2.4K minified and gzip'd isn't bad for my use cases).

aalpern:
OK, I've create a new PR (#699) which addresses the array-vaued parameter case without requiring `qs`.

PR #699, opened 2015-07-09
Encode array values as key=1&key=2&key=3 etc...

aalpern:
This partially addresses #670 for the specific case of array-valued query parameters without adding the `qs` module to the client side bundle. 

Unlike PR #698, it won't actually make the browser and node versions behave the same, since `qs` covers a LOT of different serialization possibilities, but it covers what's probably the most common case without the download hit of `qs`.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -97,9 +97,17 @@ function serialize(obj) {
   if (!isObject(obj)) return obj;
   var pairs = [];
   for (var key in obj) {
-    if (null != obj[key]) {
-      pairs.push(encodeURIComponent(key)
-        + '=' + encodeURIComponent(obj[key]));
+    var value = obj[key]
+    if (null != value) {
+      var encodedKey = encodeURIComponent(key)
+      if (Array.isArray(value)) {
+        for (var i = 0; i < value.length; i++) {
+          pairs.push(encodedKey + '=' + encodeURIComponent(value[i]))
+        }
+      } else {
+        pairs.push(encodedKey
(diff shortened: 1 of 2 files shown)


artnez:
+1

js62789:
+1

aalpern:
Hi - any chance this will ever get addressed?

pho3nixf1re:
This is a critical capability rendering this lib just about unusable in the browser. What is holding it up from merging? I would be fine with a peer dependency on 'qs' as I'm already using that library elsewhere in our app.

paranoidjk:
This is a break change, why it been released in a minor version  [1.5.0](https://github.com/visionmedia/superagent/releases/tag/v1.5.0)? 

Does superagent do not obey semver?

paranoidjk:
PS. Why not use `?cars[]=Saab&cars[]=Audi`, It's more safe.

ref https://stackoverflow.com/questions/6243051/how-to-pass-an-array-within-a-query-string

kornelski:
Sorry, I did not realize it's was breaking change. superagent intends to follow semver. The current supported version is 3, so we're past a couple of semver bumps already, so this change is going to stay.

It's intentional that `[]` is not added automatically. This syntax is an informal convention used by some server-side frameworks, **but not all**. Adding `[]` would prevent superagent from being able to send `?a=1&a=2&a=3`. OTOH you can use `.query({'a[]': [1,2,3]})` to have the `a[]` syntax, so this way superagent supports both ways.

PR #703, opened 2015-07-17
Add browser-client support js files

twincle:
Add browser-client support js files. Web developers can download js files directly for there web page ajax function


defunctzombie:
This is not needed. You can just build the bundle you need. This just bloats the repo and we will never keep it up-to-date.

PR #709, opened 2015-07-27
lib: added fake promises

matthewmueller:
This is helpful for working with co-routine control flow libraries like `co()`:

```js
var res = yield request
  .get('http://localhost:8080')
  .type('json')
```

No idea if I'm committing blasphemy in the promise world or if this will quiet the promise people down a bit. 

Don't really mind if this gets merged or not, but if people are happy with it, I'll add some tests.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -978,6 +978,20 @@ Request.prototype.end = function(fn){
 };
 
 /**
+ * Faux promise support
+ *
+ * @param {Function} fulfill
+ * @param {Function} reject
+ * @return {Request}
+ */
+
+Request.prototype.then = function (fulfill, reject) {
+  return this.end(function(err, res) {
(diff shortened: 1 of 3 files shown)


defunctzombie:
I like this compared to the alternative since it doesn't increase the file size :).

Please add tests and it can be merged.

Also need a docs update PR.

matthewmueller:
Alrighty, this should be ready for review. I'll open a PR updating the docs as well.

asaf-romano:
@defunctzombie Would you also accept an implementation of ```catch``` (in-spec sugar for ``then(null, func)```)?

matthewmueller:
I think that makes sense if people will find this as a useful compromise for promises. I don't really know, because I'm not going to be using it that way. It's not useful in the `co` generator environment though, so that's why I didn't include it.

PR #710, opened 2015-07-29
add generator docs

matthewmueller:
Okay added some docs for generators. I wasn't sure what `markdown`:

img

is being referred to here. I don't have that lib installed. But here are some docs. They'll need to be rebuilt.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -436,3 +436,14 @@ Your callback function will always be passed two arguments: error and response. 
     else if (err) {
       // all other error types we handle generically
     }
+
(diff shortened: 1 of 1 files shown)


defunctzombie:
Something about how this works via the `.then` function would help folks if they run into issues.

focusaurus:
LGTM. Should be OK to merge. No release needed.

focusaurus:
@olalonde @pornel another one we should be able to quickly review and merge.

pornel:
Thanks

PR #711, opened 2015-07-29
Fix #236 DELETE should set Content-Length to 0

### LINKED ISSUE #236: Set Content-Length header on DELETE, PUT when no entity body
Similar to https://github.com/mikeal/request/pull/41 - I'm seeing issues with an nginx server handling an API where DELETE operations to it are now returning HTTP 411 - there's a transfer-encoding header being injected somewhere along the way (not by superagent, likely nodejs itself because there's no content-length header) and it is causing severe problems with http requests.

soyuka:


test/node/delete.js
```diff
@@ -0,0 +1,21 @@
+var EventEmitter = require('events').EventEmitter
+  , request = require('../../')
```
defunctzombie: please follow existing style for source files/vars

lib/node/index.js
```diff
@@ -722,6 +722,7 @@ Request.prototype.request = function(){
   // request
   var req = this.req = mod.request(options);
   if ('HEAD' != options.method) req.setHeader('Accept-Encoding', 'gzip, deflate');
+  if ('DELETE' == options.method) req.setHeader('Content-Type', 0);
```
defunctzombie: does the client side need to be updated accordingly?

test/node/content-type.js
```diff
@@ -32,4 +32,12 @@ describe('req.set("Content-Type", contentType)', function(){
     });
   });
 
+  it('should delete with a content-length of 0', function(done){
+    request.del('http://localhost:3005/nonexistant')
+    .end(function(err, res) {
+      assert(err)
+      res.request.req._headers['content-length'].should.equal(0)
```
soyuka: Do you have a better way to check the header that has been sent?

lib/node/index.js
soyuka: What are you meaning by client? The client can sent the header he wants (see last test). Some web servers (like nginx) will deny a `DELETE` request when `Content-Length` is not set to `0`.

defunctzombie:
@soyuka the `client` aka browser version of this library

soyuka:
@defunctzombie nope, the browser should set the `Content-Length` header automatically. XMLHttpRequest isn't allowed to set it anyway. 

Could you check out my last comment : https://github.com/visionmedia/superagent/pull/711#issuecomment-126117830 ? Thanks.

soyuka:
Bump

pornel:
As far as I know HTTP RFC does not forbid DELETE requests to have a body (even GET requests can have one!), so forcing 0-length body wouldn't be appropriate.

Sorry @soyuka. That's a well-written pull request, but I think it's not a behavior the library should have.

soyuka:
@pornel maybe that was the cause:

>  A server MUST generate a Content-Length field with a value of "0" if no payload body is to be sent in the response.

https://tools.ietf.org/html/rfc7231

Thanks for the answer. Anyway, I'm almost always extending supertest/superagent for my needs. I can give an example if needed.

You may close https://github.com/visionmedia/superagent/issues/236 if it's not an issue.

pornel:
That clause is for OPTIONS (in that case I wouldn't be opposed to setting content-length if nodejs doesn't do that already).

For DELETE I see:

> A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.

PR #717, opened 2015-08-06
1. can auto post array in from-urlencode type 2. add test

yeatszhang:
this feature is really convenient


defunctzombie:
How would a dev use the API now versus before to do this?

defunctzombie:
Please clean up the commit message to just say which feature was added

yeatszhang:
I will use querystring~~

PR #719, opened 2015-08-07
Raise error if end() is called with undefined argument instead of noop-ing

dj:
I expect Request.end() to raise an error if you pass it an undefined argument. Instead, it calls a no op.

If this change is acceptable, I can add a test for it.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -896,7 +896,9 @@ Request.prototype.end = function(fn){
   var data = this._formData || this._data;
 
   // store callback
-  this._callback = fn || noop;
+  if ( fn ) {
+    this._callback = fn;
+  }
 
   // state change
   xhr.onreadystatechange = function(){


alex94puchades:
Thanks for your involvement. However, sending a request (as `request.end` does) not subscribing to the response (that is, with no callback) seems like a legit use case.

dj:
@alex94puchades, ah, this seems obvious to me now but I wasn't thinking of that before. I just had an idea though.

What if you raised the error `if ( arguments.length == 1 ) && ( typeof fn == 'undefined' )`?

However, after typing this out, I guess I could see code where you pass end a callback that only is only defined sometimes and depend on it to noop? Feel free to close this if you're convinced that passing end() an undefined argument is a valid use case as well.

I'm probably too tired to be thinking about this right now, so sorry if this is a dumb idea for some obvious reason.

focusaurus:
Not convinced this in an improvement. I recommend closing.

PR #721, opened 2015-08-10
Fixing "undefined is not a function"

nemisj:
Whenever no callback is privided to the superagent request, it crashes with `TypeError: undefined is not a function`


defunctzombie:
Don't like printing to the console directly. This spams the stdout. We could either throw or just let undefined is not a function error do what it is doing.

Issue #705, opened 2015-07-20
examples/simple-get.js url endpoint invalid


oleersoy:
$:~/Lab/superagent$ node examples/simple-get.js 

/examples/simple-get.js:11
  if (err) throw err;
                 ^
Error: Not Found
    at Request.callback (/superagent/lib/node/index.js:792:17)
    at IncomingMessage.<anonymous> (/superagent/lib/node/index.js:990:12)
    at IncomingMessage.emit (events.js:117:20)
    at _stream_readable.js:929:16
    at process._tickCallback (node.js:419:13)

reinaldo13:
Fixed and awaiting approval for PR https://github.com/visionmedia/superagent/pull/723

natarajanmca11:
I am getting the same error in post request. Details are given below.

`"superagent": "^1.8.3",`

![image](https://cloud.githubusercontent.com/assets/9244766/15352263/7426d058-1d37-11e6-9513-da73d45c9f13.png)

pornel:
This bug is about running `examples/simple-get.js`, not post requests, and not Redis.

"Not Found" error is very generic and can correctly happen in lots of situations. You should check whether you're calling correct URL and whether your server allows POST for that url.

PR #723, opened 2015-08-13
705: examples/simple-get.js url endpoint invalid

reinaldo13:
Added a valid `gist` link so `node examples/simple-get.js` can run successfully. This is to fix: https://github.com/visionmedia/superagent/issues/705

### DIFF
--- a/examples/simple-get.js
+++ b/examples/simple-get.js
@@ -5,7 +5,7 @@
 
 var request = require('..');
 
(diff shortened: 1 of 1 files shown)


focusaurus:
LGTM. Recommend merge. Doc-only update so no semver bump necessary. @pornel @olalonde please review and merge if you agree. @reinaldo13 thank you for your contribution!

Issue #724, opened 2015-08-19
isJSON heuristics override perfect MIME match


alex94puchades:
I am writing a superagent addon (not published yet) for consumption of [JSON-API](http://jsonapi.org/format/) formatted responses. This format is essentially JSON, but it dedupes repeated data in an "included" response field. The thing is, I am defining a parser for "application/vnd.api+json" as mandated by the [JSON-API](http://jsonapi.org/format/) format. However, the parser is not being called because of superagent JSON-like heuristics:

```js
// everyone wants their own white-labeled json
if (isJSON(mime)) parse = exports.parse['application/json'];

...

function isJSON(mime) {
  if (!mime) return false;
  var parts = mime.split('/');
  var type = parts[0];
  var subtype = parts[1];
  return subtype && subtype.match(/json/i);
}
```

I think the correct thing to do would be to apply heuristics ONLY when there's no perfect MIME match. Something like: `if (!parser && isJSON(mime))` would do the job and should be backward-compatible (who would expect a perfect match being overridden by a partial match, aka `/+json$/`, to be the standard behaviour?) Should be addressed also on the client-side.

alex94puchades:
Addressed by PR #725

PR #725, opened 2015-08-19
Do not apply JSON heuristics if already have a valid parser

alex94puchades:
This PR addresses #724. On a side note, code on the browser side is wildly different (this bug didn't seem to exist there).


defunctzombie:
I like it. LGTM

defunctzombie:
@alex94puchades can you please make a test for this

PR #726, opened 2015-08-19
ES6 Promise support

### LINKED ISSUE #722: The .then() function is not compatible with Promises
The faux `.then()` function added in 1.3 completely misses the point of promises and breaks even in simple cases.

```js
request.get(someCorrectUrl)
    .then(function(){return 1;})
    .then(console.log.bind(console));
```
Should print `1`, but throws an unhandled exception.

```js
request.get('fail').then(function(){});
```
> TypeError: reject is not a function

```js
request.get('fail')
    .then(undefined, function(){throw Error();})
    .catch(function(){})
```
> .then(...).catch is not a function

Please don't try to reimplement promises — just enable use of a correct library.

I suggest:

```js
Request.prototype.then = function(resolve, reject) {
  var self = this;
  return new Promise(function(resolve, reject){
    self.end(function(err, res){
      if (err) reject(err); else resolve(res);
    });
  })
  .then(resolve, reject);
};
```

The code above assumes that promises are either supported natively (already the case in majority of browsers and even Node 0.12) or polyfilled by the user (which is a common practice, e.g. ES6 transpilers do it).

pornel:
Fixes #722

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -978,16 +978,33 @@ Request.prototype.end = function(fn){
 };
 
 /**
- * Faux promise support
+ * Returns ES6 Promise if supported (or polyfilled),
+ * or fakes support by calling the callbacks and returning self.
  *
  * @param {Function} fulfill
  * @param {Function} reject
- * @return {Request}
+ * @return {Promise}
  */
 
 Request.prototype.then = function (fulfill, reject) {
+  var self = this;
+
+  if ('function' === typeof Promise) {
+    return new Promise(function(fulfill, reject){
+      self.end(function(err, res) {
+        if (err) reject(err); else fulfill(res);
+      });
+    })
+    .then(fulfill, reject);
+  }
+
+  // For backwards-compatibility with 0.13
   return this.end(function(err, res) {
-    err ? reject(err) : fulfill(res);
+    if (err) {
+      if (reject) reject(err);
+    } else {
+      if (fulfill) fulfill(res);
+    }
   });
 }
 
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -1024,16 +1024,33 @@ Request.prototype.end = function(fn){
 };
 
 /**
- * Faux promise support
+ * Returns ES6 Promise if supported (or polyfilled),
+ * or fakes support by calling the callbacks and returning self.
  *
  * @param {Function} fulfill
  * @param {Function} reject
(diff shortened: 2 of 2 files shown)


kahnjw:
This addresses my compatibility concerns by adding a thenable fallback, but is weird/unexpected from a behavioral standpoint.

Also, I'd suggest returning a promise from the `.end()` method if there is community support for it. Having the `then` method fire the request is arbitrary and weird. If the goal is to use real promises, we should use promises as they were intended - then methods should not perform actions other than to return a new promise. You could also add a new function `.endPromise()`, or something like that to address backward compatibility concerns.

pornel:
I agree returning promise from `end()` would be better.

I also think it should cache the promise, so that `var tmp = superagent.get(); tmp.then(); tmp.then()` doesn't try to launch the request again.

pornel:
BTW: the code for both node and browser happens to be identical. Do you have a preferred way for sharing/reusing it?

euwest:
>I also think it should cache the promise, so that var tmp = superagent.get(); tmp.then(); tmp.then() doesn't try to launch the request again.

Yeah, which is further reason for `end` returning the promise and it not being a part of `then`(maybe that's what you were saying?). That way you'd have:
```js
var tmp = superagent.get().end();

tmp.then(); 
tmp.then();
```

defunctzombie:
I am gonna say it again. I do not care about supporting promises in this library. The `.then` hack was to make it a simple thenable and I was fine with that. If you want a promise then use a wrapper.

Yes, we need to document the `.then` call better.

vire:
What's the status please, there is an closed PR, but in the latest version of superagent (v1.4.0 and node v0.12.7) is not working as expected (see #722).

FoxxMD:
@vire It was closed but not merged. @defunctzombie said above that he will not support A+ promises

vire:
@FoxxMD thanks for clarification. It's a pity indeed to not be complaint with some standard approach.

euwest:
>It's a pity indeed to not be comlpaint with some standard approach.

Superagent is not a promises implementation, though. Request objects just have a `then` method to make it a little easier to coerce into a promise, given an actual implementation.

niftylettuce:
Here's a promise implementation (uses ES6/ES7 async/await too, with ES6's new `fetch()` method) https://github.com/glazedio/frisbee.

PR #727, opened 2015-08-20
Added superagent-jsonapify addon to the list

alex94puchades:
Just a plugin of mine.


focusaurus:
This is just a doc update for a plugin. LGTM. Recommend merge. Does not need a release.

PR #728, opened 2015-08-20
added fields function to client.js to handle multiple fields

tarikakyol:


focusaurus:
I'd like to see both client.js and node/index.js updated to maintain API parity. I'd also like so see this implemented just by looping over object properties and dispatching to `this.field(name, value)`. Not sure silently ignoring any non-object arguments is good behavior, but whatever you propose I'd like to see the API docs clearly explain what happens if you pass a non-object argument.

tarikakyol:
commit modified as requested @focusaurus

lib/client.js
```diff
@@ -700,6 +700,33 @@ Request.prototype.query = function(val){
...
+ *   .fields({'foo', 'bar'})
+ *   .end(callback);
+ * ```
+ *
+ * @param {Object} data
+ * @returns {Request} for chaining
+ * @api public
+ */
+Request.prototype.fields = function(data){
+  if (!this._formData) this._formData = new root.FormData();
```
focusaurus: These lines duplicated code already in `.field` and should be removed.

lib/client.js
```diff
@@ -700,6 +700,33 @@ Request.prototype.query = function(val){
 };
 
 /**
+ *
+ * Write the field `name` and `val` of objects for "multipart/form-data"
+ * request bodies.
+ *
+ * ``` js
+ * request.post('/upload')
+ *   .fields({'foo', 'bar'})
```
focusaurus: This syntax is incorrect. You need a colon here instead of a comma.

focusaurus:
Still would like the docs to explain how non-objects are handled and think ignoring them is bad, fail-silently behavior likely to make debugging take much longer than necessary.

Needs at least a basic unit test to be seriously considered for merging.

@pornel what are your thoughts on this?

pornel:
We have `.set('foo','bar')` and `.set({foo:'bar'})`, so I think `.field` could be changed to behave the same.

PR #731, opened 2015-08-24
Add support for ES6 promise with fallback on es6-promise

bnadim:
End doesn't return anymore the request instance but a promise.
End still support callbacks.
End is usually the end of a call chain so i believe make it return a promise makes sense
Ex : 
```js
request.post('api/pet')
.end()
.then(function (res) { console.log(res) })
.catch(function (err) { console.log(err) })
```


defunctzombie:
Please search for other issues around this topic

Issue #635, opened 2015-04-28
Using plugin across all requests


jeffcharles:
I was looking at https://github.com/visionmedia/superagent#plugins and tried to use that to apply a plugin across multiple uses of request.

I tried something that looked similar to what was in the docs:

```
var request = require('superagent');
var prefix = require('superagent-prefix')('/static');

prefix(request);  // Prefixes *all* requests

request.get('http://localhost/foo');
```

and got this error:

```
        if (request.url[0] === '/') {
                       ^
TypeError: Cannot read property '0' of undefined
    at /Users/jcharles/projects/plugin-test/node_modules/superagent-prefix/index.js:3:24
    at Object.<anonymous> (/Users/jcharles/projects/plugin-test/index.js:4:1)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:935:3
```

I was not expecting it to fail at the prefix step since it's pretty much just code from the readme.

alex94puchades:
The example works as-is for me, could you please indicate package versions used?

jeffcharles:
Still getting the same error with 1.3.0 of superagent and 0.0.2 of superagent-prefix. I did notice that the problematic example that my code copies was removed in https://github.com/visionmedia/superagent/commit/c279f59bd4a5485a05cefca167a2d318e5591591. If that's the case, how am I supposed to apply a plugin across all uses of request without including it in each call?

alex94puchades:
Ok, now I get this error too. According to c279f59, superagent-prefix does not support global usage. The correct usage would be per-request `request.use(prefix)`. Global plugin registration is only possible atm for methods patching the superagent api, like superagent-mock does. I'll make a PR and let's see if it gets through.

DylanPiercey:
This would be extremely useful for things like prefixing, intercepts and more. I personally find single request plugins somewhat useless.

pornel:
I agree that it's useful to be able to set a default set of plugins once for all requests in an application.

However, I'm not fond of mutating a globally shared superagent module, as it could cause problems for 3rd party libraries that expected to use a vanilla module, and it creates state that can leak across unit tests.

[...]

PR #732, opened 2015-08-25
Added support for global plugin registration

alex94puchades:
This PR adds support for global plugin registration (addresses issue #635)


pornel:
Thank you for the PR. The functionality is useful, but I'm worried about exposing a shared global field. I think it's better handled by https://www.npmjs.com/package/superagent-defaults

linxixuan:
searching for this

PR #734, opened 2015-08-26
phonegap/cordova Fixes #733

### LINKED ISSUE #733: Can't use with phonegap
When trying to request a local file compiled into a phonegap/cordova application, superagent throws a `Origin is not allowed by Access-Control-Allow-Origin` error.

This can be seen with code as simple as:

```
request.get('mylocalfile.json').end((err, res) => {
    console.log(err);
});
```

Doing the same thing with vanilla JS works fine:

```
function listener() { console.log(this);}

var url = "mylocalfile.json";
var req = new XMLHttpRequest();
req.addEventListener('load', reqListener);
req.addEventListener('error', reqListener);
req.open("get", url, true);
req.send()
```

ccapndave:
Admittedly, this is a bit hacky, but it now works perfectly on Cordova/Phonegap.  I'm not totally happy about forcing it to json if the content-type is empty, but 99% of the time its going to be json so I didn't spend too long looking for a alternative solution.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -306,6 +306,10 @@ function Response(req, options) {
   // getResponseHeader still works. so we get content-type even if getting
   // other headers fails.
   this.header['content-type'] = this.xhr.getResponseHeader('content-type');
+  
+  // If we still didn't get a content-type and we are on cordova, just assume it was json (not ideal, but it usually is!)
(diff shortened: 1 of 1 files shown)


pornel:
The code *is* a bit hacky :/

Has this bug been reported to Cordova? Is that a problem they can't fix on their end?

PR #735, opened 2015-08-28
Add missing statusCode in client response

ertrzyiks:
Hey, I noticed missing statusCode on client side.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -394,7 +394,7 @@ Response.prototype.setStatusProperties = function(status){
   var type = status / 100 | 0;
 
   // status / class
-  this.status = status;
+  this.status = this.statusCode = status;
   this.statusType = type;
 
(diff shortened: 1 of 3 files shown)


defunctzombie:
Where is it documented that we use statusCode?

ertrzyiks:
We use superagent in isomorphic app and we experienced inconsistency in client and server response object. I don't see any documenation about statusCode indeed and I even tried to remove it, but testsuite failed without this property. 

I've added this field to introduce consistency, but similar effect can be done by removing statusCode from node version.

defunctzombie:
Gotcha, yea I can see in the changelog and code that it was added to mirror the node response object. Don't want to undo that change so this seems like a reasonable patch-up. It will remain undocumented but can be something that is there for use since the node version already had it.

PR #739, opened 2015-09-04
Don't modify the query string, but preserve its encoding

RubenVerborgh:
Currently, if you pass an URL as string, the **query string is parsed and regenerated**.

On the one hand, this **introduces overhead**, because such parsing and generation is not necessary.

More importantly, however, it **destroys the original encoding of the passed URL**. Therefore, it is currently impossible to use superagent with servers that are sensitive to encoding. In particular, we cannot use superagent (and thus supertest) to verify whether a server correctly reacts to different encodings. For instance, if we want to try `http://example.org/a?q=(b)` and `http://example.org/a?q=%28b%29`, superagent would in both cases send `http://example.org/a?q=(b)` to the server.

This commit preserves the original URL encoding by adding the query string to `qsRaw` rather than `qs`.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -650,6 +650,7 @@ Request.prototype.redirect = function(res){
   this._redirectList.push(url);
   this.emit('redirect', res);
   this.qs = {};
+  this.qsRaw = [];
   this.set(headers);
   this.end(this._callback);
   return this;
@@ -716,7 +717,7 @@ Request.prototype.request = function(){
 
   // default to http://
   if (0 != url.indexOf('http')) url = 'http://' + url;
-  url = parse(url, true);
+  url = parse(url);
 
   // options
   options.method = this.method;
@@ -756,7 +757,8 @@ Request.prototype.request = function(){
   }
 
   // query
-  this.query(url.query);
+  if (url.search)
+    this.query(url.search.substr(1));
 
   // add cookies
   if (this.cookies) req.setHeader('Cookie', this.cookies);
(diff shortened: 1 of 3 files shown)


lib/node/index.js
```diff
@@ -640,6 +640,7 @@ Request.prototype.redirect = function(res){
   this._redirectList.push(url);
   this.emit('redirect', res);
   this.qs = {};
+  this.qsRaw = [];
```
RubenVerborgh: This actually was a bug in the original implementation of redirection: only the key/value part of the query string was cleared (`qs`), but not the raw query string parts (`qsRaw`). This behavior surfaced because this commit now stores the original query string in `qsRaw`, and an existing unit test verifies whether the query string is cleared on redirect.

lib/node/index.js
```diff
@@ -706,7 +707,7 @@ Request.prototype.request = function(){
 
   // default to http://
   if (0 != url.indexOf('http')) url = 'http://' + url;
-  url = parse(url, true);
+  url = parse(url);
```
RubenVerborgh: We don't need to parse the URL's query string anymore, as we use it unmodified.

lib/node/index.js
```diff
@@ -743,7 +744,8 @@ Request.prototype.request = function(){
   }
 
   // query
-  this.query(url.query);
+  if (url.search)
+    this.query(url.search.substr(1));
```
RubenVerborgh: If there is a query string, pass it unmodified (skipping the leading `?` character).

pornel:
Thanks for the PR. Sorry for the late response. The change looks OK. Could you rebase it?

RubenVerborgh:
@pornel Rebased and ready to merge.

pornel:
Thank you.

PR #740, opened 2015-09-11
Added Request#parse method to client build

alex94puchades:
Allows browser clients to override response parser (already allowed in node builds) (as part of #690)

PR #742, opened 2015-09-15
Allow custom parser in Request [was: Fixes #740]

alex94puchades:
Added parse method to Request class instead of Response

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -640,6 +640,20 @@ Request.prototype.type = function(type){
 };
 
 /**
+ * Force given parser
+ * 
+ * Sets the body parser no matter type.
+ * 
+ * @param {Function}
+ * @api public
+ */
+
+Request.prototype.parse = function(fn){
+  this._parser = fn;
+  return this;
+};
+
+/**
  * Set Accept to `type`, mapping values from `request.types`.
  *
  * Examples:
@@ -961,7 +975,7 @@ Request.prototype.end = function(fn){
   if ('GET' != this.method && 'HEAD' != this.method && 'string' != typeof data && !isHost(data)) {
     // serialize stuff
     var contentType = this.getHeader('Content-Type');
-    var serialize = request.serialize[contentType ? contentType.split(';')[0] : ''];
+    var serialize = this._parser || request.serialize[contentType ? contentType.split(';')[0] : ''];
     if (serialize) data = serialize(data);
   }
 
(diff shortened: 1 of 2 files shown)


defunctzombie:
What is the issue? Please make a better commit message. Was the previous commit not tested? Please add tests for this now, clearly the previous commit was bad and I should have been more thorough in reviewing it.

lib/client.js
```diff
@@ -980,7 +980,8 @@ Request.prototype.end = function(fn){
   if ('GET' != this.method && 'HEAD' != this.method && 'string' != typeof data && !isHost(data)) {
     // serialize stuff
     var contentType = this.getHeader('Content-Type');
-    var serialize = request.serialize[contentType ? contentType.split(';')[0] : ''];
+    var contentParser = request.serialize[contentType ? contentType.split(';')[0] : ''];
```
defunctzombie: `this._parser || ` can just prefix this to avoid the lookup if we have a parser

alex94puchades:
@defunctzombie Sorry, really, it was my bad. This is what happens when "fast" gets in the place of "good". I am really sorry. The issue with the last commit was that it added the ability to override a parser for the Response, when what I wanted to was to add that ability to Request. It WAS tested, it was tested not to break anything that wasn't broken, and I was adding unit tests sometime in the future. Everything I said still applies: no backward-incompatible changes, no broken tests, it was just a typo.

defunctzombie:
Can you add a client test for these changes so it doesn't break in the
future and we know it works plz :)

lib/client.js
alex94puchades: Is that a real issue? Sorry, I was just trying to keep within the 80 character line width

defunctzombie: Nope, just a suggestion.

alex94puchades:
Wow, seems like this is gonna take a little more time than I thought : /

alex94puchades:
I fear this is gonna take me some time : /

alex94puchades:
@defunctzombie pffew, all tests pass now :)

alex94puchades:
@defunctzombie Ready to be merged at last! ^^

defunctzombie:
please remove the dummy commit and squash remaining commits into a single one with a commit message describing the change

alex94puchades:
Done

defunctzombie:
@alex94puchades what's with the "fix detection of root object" commit?

alex94puchades:
Sorry, I think it got included while squashing commits, I hope everything is fine now

defunctzombie:
:+1:

Issue #605, opened 2015-03-27
Favor request.delete() over request.del()


dennishall1:
Request itself has to manage del as a special case of the REST verbs.  Most code that uses request also has to treat delete specially, as del.   Request could benefit from treating DELETE the same as all other verbs, and map it to request.delete instead of request.del.  (This update to request would align with how express treats delete:  http://expressjs.com/api.html#app.delete.method)

```
methods.forEach(function(method){
  var name = 'delete' == method ? 'del' : method;
```

rase-:
I do think parity with `express` would make a lot of sense, not to mention consistency of verb method names, and would be in favor of this. 

cc @defunctzombie for an opinion

I wonder if there is another reason for missing `.delete` other than reserved names being generally considered bad practice for property names.

defunctzombie:
I have been bitten by this a few times but the moment we change it, I am almost 100% certain someone will complain and say it doesn't work in `use strict` or something. As long as it works under `use strict` then I have no problem with making an alias.

shesek:
It should be fine with strict mode as long as `obj['delete']` is used rather than `obj.delete`.

dennishall1:
Shouldn't be anything wrong with `obj.delete`.  Strict mode should only disallow an attempt to set a value to a _global variable_ named `delete`, but this is not a global variable, it is an object's method name.

Seems fine:
`(function(){ "use strict";  var o = {};  o.delete = "hello"; return o; })();`
..."Use strict" does not appear to prevent setting an Object property named delete.

The only potential gotchya would be if the language ever decided to add an un-overridable `delete` method or property to the Object prototype.  `Object :: deleteProperty` is in the ES6 draft.

defunctzombie:
Would review a PR for this.

iMerica:
This has been closed, yet `.delete()` is still not supported. What was the final resolution?

shesek:
@iMerica This issue is still opened, the "closed" you're seeing is from the referenced issue at [borisyankov/DefinitelyTyped](https://github.com/borisyankov/DefinitelyTyped).

focusaurus:
Fixed by #743

PR #743, opened 2015-09-17
605: adds delete verb support and maintains del verb support

aaroncrows:
Addresses #605. This maintains support for 'del' while adding 'delete' so that it's not a breaking change. On the client side it maps the same function that was used for 'del' to both verbs. On the server side it adds 'del' to the list of methods if it isn't there and maps them both to the 'DELETE' request. Added a test for the client that's the same as del. Doesn't seem like this should be an issue with 'use strict' but I could easily add a test for that since it can be scoped to a function.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1076,11 +1076,14 @@ request.head = function(url, data, fn){
  * @api public
  */
 
-request.del = function(url, fn){
+function del(url, fn){
   var req = request('DELETE', url);
   if (fn) req.end(fn);
   return req;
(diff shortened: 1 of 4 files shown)


focusaurus:
LGTM. API addition so semver minor bump. @pornel @olalonde please review and merge if you approve. @aaroncrows thanks for your good work on this one!

pornel:
We don't support IE8, and ES5 allows reserved words as property names, so it's all good.

PR #750, opened 2015-10-01
confusing behaviour of buffer setting for custom parsers

Cwallice:
Looks like custom parsing didn’t propagate correct response object back

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -948,7 +948,7 @@ Request.prototype.end = function(fn){
     // explicit parser
     if (parser) parse = parser;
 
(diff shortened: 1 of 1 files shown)


Cwallice:
I faced this problem while doing proxy responses with "application\yaml" content type
Since proxy our engine is very generic it really doesn't care about possible return content type before response is actually received, since that I can't set parse function and while buffer option set to true is really a solution it still bring confusing because it doesn't come along with parsing settings logic

pornel:
I think the buffering situation was fixed in 2.3

PR #758, opened 2015-10-07
Fix ie11 sends 'undefined' string if data was undefined. Issue #632

### LINKED ISSUE #632: An empty body with PUT on ie11 sends undefined down the wire.
The following code sends content-length: 0 and no body on FF and Chrome, but on IE11 the browser sends "undefined" with content-length: 9.

```
superagent('PUT', 'example.com').set('API-Key', 'foobar').end(function(err, res){
   // handle response
});
```

I used superagent 1.2.0, IE11 via modern.ie and tcpdump to verify this. The request was a CORS request with an initial OPTIONS call identical to Chrome and FF.

Can anyone reproduce the error?

vadimgoncharov:
Fixed #632

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -992,6 +992,9 @@ Request.prototype.end = function(fn){
 
   // send stuff
   this.emit('request', this);
+
+  // IE11 xhr.send(undefined) sends 'undefined' string as POST payload (instead of nothing)
+  // We need null here if data is undefined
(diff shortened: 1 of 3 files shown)


focusaurus:
LGTM. Bugfix so semver patch. @vadimgoncharov if you feel inclined to add a unit test, we'd love that. @pornel @olalonde please review and merge if you agree.

vadimgoncharov:
@focusaurus done.

PR #759, opened 2015-10-07
Split url-encoded parameters in the standard way

shurizzle:


pornel:
Thank you for the PR. It's a useful change.

1. I think the code could be simplified by checking `parts.length` from `parts = split('=')`.

2. The change from `"undefined"` to `undefined` is good, but it may be a breaking change, so it'll need a major version bump.

PR #761, opened 2015-10-08
ensure query string is appended when doing stream.pipe(request)

keeth:
Query string is missing from request path when using stream.pipe(request)

node repl test case:
```
var express = require('express'), fs = require('fs'), request = require('superagent');
express().put('/', function(req,res){ console.log("url=" + req.url) }).listen(3006);
fs.createReadStream('README.md').pipe(request.put('http://localhost:3006/?foo=bar'));
```
expected output: url=/?foo=bar
actual output: url=/

The problem: the query is appended to req.path in end() which is run after the 
headers have been sent when the request is treated as a writable stream, so the
path sent to the server is always missing the query.

My attempt at a fix: the first time write() is called, set a flag indicating
that the request is being treated as a writable stream.  If this is the case we 
need to append the query string to req.path before the first write() causes 
the headers to be sent.

I'm not super familiar with this library, but this seemed like decent
fix that would solve the problem while keeping the existing behavior of 
passing any query-related error to the end() callback, for the more common
case of calling req.end() rather than stream.pipe(req) to send the request.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -132,6 +132,7 @@ function Request(method, url) {
   this.qs = {};
   this.qsRaw = [];
   this._redirectList = [];
+  this._streamRequest = false;
   this.on('end', this.clearTimeout.bind(this));
 }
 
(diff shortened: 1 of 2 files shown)


focusaurus:
Seems reasonable to me, but also probably worth some detailed review. @pornel @olalonde thoughts?

pornel:
It's an interesting bug, and potentially very annoying gotcha, so we definitely should fix it. I'm not familiar with the code enough myself to be sure what would be an ideal solution.

However, flags like that are a shared mutable state, and code that modifies query string is now duplicated. So while this fixes this particular bug, I'm afraid it could be easily broken later (e.g. if qs serialization is updated in end() but not in write()).

pornel:
OK, it's better to fix this. We'll worry about a stateful property later.

PR #769, opened 2015-10-14
Fix to avoid throwing errors on aborted requests

gjurgens:
I'm downloading large sets of json files using superagent from a slow server. In some cases, I need to abort the request using `.abort()` method.
When the json is in the middle of the transfer, I get a syntax error when superagent tries to parse the partially transfered json string. 
I think that on aborted requests, the parser shouldn't throw any error.

Thanks for the good work, and let me know if I can help fixing this.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -952,7 +952,7 @@ Request.prototype.end = function(fn){
     if (parse) {
       try {
         parse(res, function(err, obj){
-          if (err) self.callback(err);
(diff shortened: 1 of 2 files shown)


focusaurus:
LGTM. Recommend merge. Semver patch vs minor is debatable but I guess the conservative path would be to consider this a semver minor change.

ChrisCinelli:
+1

Issue #516, opened 2014-12-20
Add progress event on node


mauricesvay:
The `progress` is only triggered on the browser version of the lib. It would be nice to have it on the node side too.

olalonde:
+1. Has this been implemented yet?

olalonde:
I started implementing this but got stuck here because writable streams don't have a 'data' event:

```javascript
  // See http://dev.w3.org/2006/webapi/progress/Progress.html
  req.on('data', function () {
    var length = req.getHeader('Content-Length');
    var lengthComputable = !!length;
    var loaded = 0;
    var total = lengthComputable ? length : 0;
    return function (data) {
      loaded += data.length;
      self.emit('progress', {
        lengthComputable: lengthComputable,
        loaded: loaded,
        total: total
      });
    };
  }());
```

I tried googling around but couldn't find any helpful info on tracking upload progress with nodejs' http module.

olalonde:
Ok, so I implemented this on `form-data` uploads only (e.g. when using .attach() or .field()).

shishirx34:
This is exactly what I was looking for, @olalonde is there an ETA on when #771 PR would be merged?

pornel:
I have concerns about the PR in the current state, and I'm waiting for a response from @olalonde, but in general I agree that's a good thing to implement.

olalonde:
PR merged though it would still be nice to have 

1) progress events on the incoming side (when downloading). (and a way to distinguish them)

2) currently, the progress event only works when using the `attach()` or `field()` APIs (anything that uses form data). It would be nice to also have it on regular requests or requests that are piped to from a readable stream.

pornel:
Agreed.

PR #771, opened 2015-10-20
Added 'progress' event on file/form upload in Node

olalonde:
See issue #516

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -1014,7 +1014,24 @@ Request.prototype.end = function(fn){
         req.setHeader('Content-Length', length);
       }
 
-      formData.pipe(req);
+      var getProgressMonitor = function () {
+        var lengthComputable = true;
+        var total = req.getHeader('Content-Length');
+        var loaded = 0;
+
+        var progress = new Stream.Transform();
+        progress._transform = function (chunk, encoding, cb) {
+          loaded += chunk.length;
+          self.emit('progress', {
+            lengthComputable: lengthComputable,
+            loaded: loaded,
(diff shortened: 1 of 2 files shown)


olalonde:
Hmmm any feedback? Is this project still maintained?

focusaurus:
@olalonde see #785 for updates on transitions to new maintainers.

LGTM. Recommend merge and semver minor bump.

pornel:
How do you distinguish between upload and download progress? Won't it confuse existing code that posts something and then expects to track *download* progress?

I don't see `lengthComputable` changed anywhere. 

Are you certain Content-Length is always available?

olalonde:
> How do you distinguish between upload and download progress? Won't it confuse existing code that posts something and then expects to track download progress?

Curiously, the browser implementation doesn't seem to distinguish between upload/download: https://github.com/visionmedia/superagent/blob/master/lib/client.js#L974 

Maybe we should address this first?

> I don't see lengthComputable changed anywhere.

Yes it's not changed. I just put it there so that the event object somewhat resembles what browser progress events: http://dev.w3.org/2006/webapi/progress/Progress.html

AFAIK, Content-Length is always available when using form data. There is actually a [TODO](https://github.com/olalonde/superagent/blob/node-progress-event/lib/node/index.js#L1010) which explicitly states that "unknown length" uploads aren't supported yet.

pornel:
Great, thank you!

shishirx34:
This is great. Thank you for getting this merged in. @pornel, any idea when the next npm update will be available for these changes on superagent?

PR #782, opened 2015-11-13
Fixed example in the documentation of the first page. Fixes #772

### LINKED ISSUE #772: first documentation example has broken error handling
The very first example at http://visionmedia.github.io/superagent/ currently shows

```
 request
   .post('/api/pet')
   .send({ name: 'Manny', species: 'cat' })
   .set('X-API-Key', 'foobar')
   .set('Accept', 'application/json')
   .end(function(err, res){
     if (res.ok) {
       alert('yay got ' + JSON.stringify(res.body));
     } else {
       alert('Oh no! error ' + res.text);
     }
   });
```

However, as noted in #760, this pattern of result checking will fail unexpectedly if there is an error in the callback passed to `.end()`. In that case, `res` is undefined and in my browser I get `Uncaught TypeError: Cannot read property 'ok' of undefined` in the line `if (res.ok) {`.

Therefore, I would suggest to update the documentation to properly perform error checking (or at least to remove the broken pseudo error check).

kfeyrer:
Solution by @astraw

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -9,10 +9,10 @@
        .set('X-API-Key', 'foobar')
        .set('Accept', 'application/json')
        .end(function(err, res){
-         if (res.ok) {
-           alert('yay got ' + JSON.stringify(res.body));
+         if (err || !res.ok) {
+           alert('Oh no! error');
          } else {
-           alert('Oh no! error ' + res.text);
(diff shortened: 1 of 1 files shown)


focusaurus:
LGTM. doc-only fix so no release needed. Recommend merge.

olalonde:
LGTM

PR #789, opened 2015-11-24
Added Bower manifest

johnnyfreeman:
From the conversations I found in #315 and #387 (and a few other issues), it seems like you guys are ok with a minimal `bower.json` file, but no one has met your requirements for that yet.

I found the [current spec for bower.json](https://github.com/bower/spec/blob/master/json.md) from the Bower team. `name` is the only required field. `main` is useful for build scripts.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,4 @@
+{
+  "name": "superagent",
+  "main": "lib/client.js"
+}


pornel:
I'm not a fan of bower, but if that's all that's needed for it then I don't mind. @focusaurus?

focusaurus:
Also not a fan of bower. I'm also not sure this will in fact work when installed via bower but I have no strong objections to merging this, especially if others in the community will shoulder the maintenance burden.

johnnyfreeman:
Out of curiosity, what do you use instead of Bower? I'm hearing this comment more and more and know there must be a reason even though it's worked well for me (so far).

focusaurus:
I vastly prefer npm for both node and browser development (with browserify). I resort to bower only when some library I need is bower-only and I can find no suitable alternative on npm, which is trending toward never on a steep trajectory.

pornel:
OK, let's give Bower a chance

PR #798, opened 2015-11-27
Avoid the error which is emitted from 'socket hang up' to cause the fn undefined error on JS runtime.

leonli:
Avoid the error which is emitted from 'socket hang up' to cause the `fn undefined error` on JS runtime.

while event `req.error` is emitted:

```
req.on('error', function(err){
    // flag abortion here for out timeouts
    // because node will emit a faux-error "socket hang up"
    // when request is aborted before a connection is made
    if (self._aborted) return;
    self.callback(err);
  });
```

because only `prototype.end()` and `prototype.redirect()` had assigned the value to `this._callback`, so if the `req.on('error')` event is fired -> `self.callback(error)`, it will cause `prototype.callback()` to throw the `undefined` runtime error.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -764,7 +764,8 @@ Request.prototype.request = function(){
  */
 
 Request.prototype.callback = function(err, res){
-  var fn = this._callback;
+  // Avoid the error which is omitted from 'socket hang up' to cause the fn undefined error on JS runtime.
+  var fn = this._callback || noop; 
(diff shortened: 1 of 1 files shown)


pornel:
I'm noticing a pattern of errors with implicit expectations of state change on end(), so eventually we should look into making it an explicit state machine to solve all such cases.

But this as a quick fix seems sensible.

PR #812, opened 2015-12-14
merge gh-pages back into master

focusaurus:
OK, gh-pages has had master merged in so it's close to master. This PR will get the doc fixes around `req().get().send()` merged in.


pornel:
*into* master? Eek. I'm not sure about that.

pornel:
My impression is that the gh-pages branch is messier and has auto-generated files. That's fine when it's used to power the website behind the scenes, but not so good in the visible part of the repository.

focusaurus:
OK let me rework my changes and PR them directly into master then, and from them we can merge into gh-pages.

PR #817, opened 2015-12-17
Share more tests between node and browser

pornel:
There's a shared test suite in `basic.js`, but it seems small compared to node-specific and browser-specific test suites.

That worries me, because in practice it means that very little of the API compatibility between node and browser is enforced automatically.

Here's a first step to share most of the tests between both implementations.


focusaurus:
Great. I've similarly been thinking about whether a BaseRequest common prototype should be extracted that extends EventEmitter and has all the copypasta between node/client code. May need some tooling updates to keep the jsdoc convenient when we generate the HTML docs. I'll wait until we get the test passing to merge this change, but otherwise +1.

PR #818, opened 2015-12-18
Move tests

pornel:
Another batch. 

It has uncovered some incompatibilities, which for now I've disabled in browser with `if (doesntWorkInBrowserYet) {}`


focusaurus:
LGTM

PR #819, opened 2015-12-18
Don't let request error override responses

clayreimann:
Talking to certain servers I was seeing the following error, even when the response was 200 OK

```
Error: read ECONNRESET
    at exports._errnoException (util.js:855:11)
    at TCP.onread (net.js:544:26)
```


clayreimann:
How can I kick off the Sauce tests again? (and why are they always so flakey?)

pornel:
I've restarted tests (you can do that by clicking details, then failed test, and a refresh button in top right)

focusaurus:
LGTM

PR #821, opened 2015-12-22
fix `delete` breaking on ie8

kenjiokabe:
Please check if mergeable.

I want to use this nice library on Ie8 with no error...
(I know it supports `>=ie9`)

Thank you.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1126,8 +1126,8 @@ function del(url, fn){
   return req;
 };
 
-request.del = del;
-request.delete = del;
+request['del'] = del;
+request['delete'] = del;
 
 /**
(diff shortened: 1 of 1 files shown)


pornel:
Does it work apart from this?

kenjiokabe:
@pornel 
Yes, it works!

cf. 
http://stackoverflow.com/questions/26479123/angularjs-http-delete-breaks-on-ie8

kenjiokabe:
@pornel 
Thank you for merging!

Could you please update npm ver. `1.6.1` -> `1.6.2` ?

lbeschastny:
:+1: It would be very nice to have this PR released

Issue #828, opened 2015-12-29
can not change request url after set header


vicanso:
suppose that I have two plugins, one for add host to request url, the other for add uuid to http header.

```js
var request = require('..');

var url = '/reinaldo13/cdbb4d663ba23410a77b/raw/0345267767d50790051951ddc460e2699649de2b/it-works.txt';

function addHost(request){
  request.url = 'https://gist.githubusercontent.com' + request.url;
  return request;
}

function addUUID(request) {
  request.set('X-UUID', Date.now());
  return request;
}

request.get(url)
  .use(addHost)
  .use(addUUID)
  .end(function(err, res){
    if (err) throw err;
    console.log(res.text);
  });

```

The code above executes exactly as what I want at first;  however, if I change the use order of plugins, it doesn't work.

```js
request.get(url)
  .use(addUUID)
  .use(addHost)
  .end(function(err, res){
    if (err) throw err;
    console.log(res.text);
  });
```

I found that when `Request.prototype.set` function is called, it will call `Request.prototype.request` to create an `http[s] request`, then `the request url` can not be changed any more.

I would like to ask whether  `Request.prototype.set` can be modified as follows:

- save the header in a temp object, like `this.header`

- when call the `Request.prototype.end` function, set the header for `http[s] request`

ps: I found that the superagent for browser acts as the description above.

vicanso:
fixed

PR #829, opened 2015-12-30
change set header function, for issue #828

vicanso:
change set header function, not call this.request() until call this.end()


pornel:
LGTM. @focusaurus?

Issue #205, opened 2013-04-11
No way to set rejectUnauthorized option of underlying node https request


georgesnelling:
TJ, for https requests, node core changed the default value of the https.request option rejectUnauthorized from false to true. 

8.22:  http://nodejs.org/docs/v0.8.22/api/https.html#https_https_request_options_callback

10.4:  http://nodejs.org/docs/v0.10.4/api/https.html#https_https_request_options_callback

This means that if a node 0.8.x client sends an https request to a server with a self-signed certificate it will succeed.  A  node 0.10.x client sending the same request to the same server will fail with Error: DEPTH_ZERO_SELF_SIGNED_CERT.  

While I agree with the change in default, we need to be able to set this bit, (similar to curl's --insecure option) in our request lib in order for our tests to hit https servers using self-signed certs.  Mikeal's request lib exposes a strictSSL option which maps to the underlying node option, but I can't find a way to twiddle that bit from superagent.   

I hate to clutter superagent's clean api, but perhaps an .options({}) method, passed through to node's request constructor, could give those who cared full access to all the underlying knobs.  There's probably a cleaner way... 

Thanks.

devangpaliwal:
process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0; Worked!!

mikemaccana:
Is it currently possible to pass `rejectUnauthorized` in options? Reading some of the linked comments suggests it is, but there's nothing mentioned in the documentation and it doesn't seem to work when I try.
For my own use case  `process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;` is dangerous.

pornel:
We don't have a solution to this yet. 

* The general idea is to have `.tls()` method with TLS-related options, but the current pull requests simply allow setting of *any* agent option, and I'm not comfortable with having an "anything goes" unlimited API.

* `rejectUnauthorized` is not scary enough. It should be `THIS_IS_TOTALLY_INSECURE_AND_WILL_EAT_BABIES = true` or such.

mikemaccana:
Thanks for the quick answer. Good point, node itself doesn't name the option very well, I've asked what people think about renaming it: https://github.com/nodejs/node/issues/5258  

In my case (I want to check an existing TLS setup so I can recommend how to fix it) I ended up just checking for `err.code === 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'` which was sufficient for my needs.

harshithjv:
[...]

Issue #188, opened 2013-03-12
ability to set rejectUnauthorized to false


andrewrk:
as of node v0.10, if you try to access an SSL endpoint which is not in node's default CA list, an error will be thrown. There should be a way to set an option so that rejectUnauthorized is set to false.

See http://nodejs.org/docs/latest/api/https.html#https_https_request_options_callback

defunctzombie:
Closing issue since there is an ENV var workaround. I don't think we have specific node options.

bryanlarsen:
An example of where this would be useful and the environment variable workaround is with webhooks.

I'm setting up webhooks in my app, and I'm copying the github API.   They have a configuration flag called 'insecure_ssl'.

yuyudhan:
@defunctzombie 

` process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0" `
This sets SSL off for all http requests in the process. Shutting off all together would be drastic in production use.

Do you know how to set it off for one particular request. Anyone??

pornel:
I don't recommend disabling HTTPS security. If you can't get a trusted certificate, then you can create your own CA and set it as trusted in superagent using `.ca()`.

http://stackoverflow.com/a/24749608/27009

rusekr:
@pornel 
We have system in production that communicates with many external systems on production and test basises (we integrating with people, other people integriate and test integration with us). Recently maintainers of one of this external systems forgot to update certificate and now part of important for us API do not work. While they requesting new cert (2-3 weeks) we set temporary  rejectUnauthorized: false for only requests to this API with checking while it is updated to revert option to true. 

But some tests with requests to this API written with chai-http which uses superagent and while i can  suppress errors for native nodejs request with rejectUnauthorized, for chai-http i can`t.

Setting process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0" kills security of many other requests even if i will be reverted in in response callback.

What to do? Change testing framework because of this small but important in real life option?

pornel:
Semi-serious, but you could put the broken-SSL endpoint behind CloudFlare CDN. They don't care about enforcing end-to-end security, but do enough to make superagent, etc. to believe it's secure.

[...]

Issue #737, opened 2015-08-31
Https with certificate


rsilvestre:
Hi,

It's more a question than a bug.

I'm using _sailsjs_ for a project and i'd like to use a extension to sails named : _sails-rest_. The next version of this extension [v0.1.0-waterline-v0.10](https://github.com/zohararad/sails-rest/tree/v0.1.0-waterline-v0.10) will use _superagent_. It look than **_superagent_ don't support https with certificate authentification**. I had a [discussion](https://github.com/zohararad/sails-rest/issues/63) with the _[Zohararad](https://github.com/zohararad)_ who developed the _sailsjs_ extension, but he asked me to check with you if you can give more information about that issue.

In my project I need to use certificate to authenticate on a API with _sailsjs_. Previously I developed a simple _express_ project and used [request](https://www.npmjs.com/package/request) with a certificate. It worked fine. So why this feature is not implemented in _superagent_, **did you expect to integrate a certificate based authentication**, if it is developed **could you give an example** on how to use it?

Regards,

timon:
Hello,

same question here.

pornel:
Would this help? #681

timon:
@pornel Thanks, looks promising!
I'll give it a try.

timon:
@pornel exactly what I looked for.

titanism:
Client certificates are supported with .ca(), .key(), .cert(), or .pfx(), including agents. The feature landed in #1057/v5.1.1; closing as resolved.

v10.4.0 is available on npm: https://www.npmjs.com/package/superagent/v/10.4.0
GitHub release: https://github.com/forwardemail/superagent/releases/tag/v10.4.0

Consider supporting our efforts to maintain this project by using Forward Email at https://forwardemail.net.  We run the only 100% open source and 3rd party audited email service in the world.

Issue #197, opened 2013-03-29
Ability to set Certificate Authority for a request or agent.


RGBboy:
As of Node 0.10.0, rejectUnauthorized default value has been set to true meaning that tests that use HTTPS locally will now fail where superagent is used.

I suggest add the ability to set the options.ca so that self signed certificates will work with superagent.

The option `ca` in the https module accepts an authority certificate or array of authority certificates to check the remote host against.

See http://nodejs.org/api/https.html#https_https_request_options_callback

PR #832, opened 2016-01-05
allow rejectUnauthorized to be set to false for https request

Whoaa512:
use `.disableStrictSSL()`, otherwise defaults to true

This is a valid use case, because simply ignoring security for the entire process with `process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0` is unacceptable, and should be allowed to be set on a per request basis.

Related: https://github.com/visionmedia/superagent/issues/188 and https://github.com/visionmedia/superagent/issues/205

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -691,6 +691,18 @@ Request.prototype.ca = function(cert){
 };
 
 /**
+ * Set the `rejectUnauthorized` option for underlying https request options.
+ *
+ * @return {Request} for chaining
+ * @api public
+ */
+
+Request.prototype.disableStrictSSL = function(){
+  this._rejectUnauthorized = false;
+  return this;
+};
+
+/**
  * Allow for extension
  */
 
@@ -725,6 +737,7 @@ Request.prototype.request = function(){
   options.host = url.hostname;
   options.ca = this._ca;
   options.agent = this._agent;
+  options.rejectUnauthorized = this._rejectUnauthorized;
 
   // initiate request
   var mod = exports.protocols[url.protocol];
--- a/test/node/https.js
+++ b/test/node/https.js
@@ -35,6 +35,17 @@ describe('https', function(){
         done();
       });
     });
+
+    it('should give a good response without strict SSL', function(done){
+      request
+      .get('https://localhost:8443/')
+      .disableStrictSSL()
+      .end(function(err, res){
+        assert(res.ok);
+        assert('Safe and secure!' === res.text);
+        done();
+      });
+    });
   });
 
   describe('.agent', function () {


pornel:
Thank you for the PR.

What kind of problem are you trying to solve? If the server has a self-signed certificate, wouldn't it be safer to have an option whitelist that certificate or its self-signed CA?

If it's aimed for testing on localhost, should it be limited to local domains only? Or work only if `NODE_ENV=development`?

Is the name consistent with any existing node API? "SSL" is an old name for TLS, and I'd prefer a more dramatic name for this option, as it's not just disabling "strictness", but completely disables all protection against MITM attackers decrypting and manipulating the content.

There were other requests for generic TLS options, #681 #737 #685 #197. Do you have suggestion how API for other TLS options should look like?

Whoaa512:
The problem in question is for making a request to an internal service with a self-signed cert, which unfortunately my IT department won't grant me access. This also rules out the dev only restriction.

I actually like the design from #681 more than my own and will continue my comments there and close this PR.

lgreenlee:
I have this problem as well. I really like your library but it is unusable without this option. I have an environment where the hosts are trusted but I need confidentiality for their communications. The lack of this option prevents me from securing my environment at all. I also think that you don't really need to worry about screening the options. Anyone looking to set TLS options probably has some idea of what they are up to. These options are also well documented for Node.js. As Patrick Bruel once said, "Let the people do what they want"! So I'd rather see the .tls style implemented. #681 #926

pornel:
Note that with `rejectUnauthorized:false` in Node *allows MITM attacks*, which means an attacker can decrypt and maliciously modify your traffic. 

So if you have hosts that need confidentiality of communication *you must not* use `rejectUnauthorized:false`.

It's not "just a bit less secure". It's plainly insecure, because your data will be "securely" sent to anybody anywhere who intercepts it, including attackers. 

Currently in superagent you can secure communication with a host that has a self-signed certificate by using the `.ca()` method. Create your own CA and issue your own certificates using it (e.g. Charles proxy can do it for you. In OS X keychain app has a "wizard" to generate these as well), and set that CA's cert as trusted in superagent.

lgreenlee:
My comments here are limited to securing service to service communication on a "quasi-private" e.g. Cloud network or Corporate network. External User to System communications should avail themselves of the full set of features of the CA infrastructure.

The underlying issue is that we use certificates to provide confidentiality _and_ identity. These actually need to be two distinct mechanisms which they are not always today. In my opinion, ensuring confidentiality is the most important aspect of securing endpoints followed by establishing identity. There are also a range of methods for identity management that go beyond the features embedded in the certificate and the CA system.

I'd also suggest that the identity management aspect of certificates for any important application is basically a failure. See [the Lenovo superfish example (arstechnica.com)](http://arstechnica.com/security/2015/02/lenovo-pcs-ship-with-man-in-the-middle-adware-that-breaks-https-connections/) (which is basically a man in the middle attack) and  [Google for example (google.com)] (https://security.googleblog.com/2015/03/maintaining-digital-certificate-security.html).

Fundamentally, the use of rejectUnauthorized does not compromise the channel's ability to provide confidential communication. It does defeat the identity management aspects of the certificate infrastructure and in some scenarios this is perfectly fine.

[Easy RSA (github.com)] (https://github.com/OpenVPN/easy-rsa) is an easy way to setup a CA that can be used in a number of different environments.

pornel:
If you say CA are rubbish, you still need to distinguish between two very different scenarios:

1. You don't verify identity at all, and you're happy to send the data to an attacker presenting an untrusted certificate

2. You verify identity using your own CA or altertnative method of certificate validation to ensure you reject attackers' unauthorize certificates

Currently superagent supports option 2, but does not allow the insecure and irresponsible option 1.

lgreenlee:
Thank you for your input. PKI is still evolving and is not perfect, but it is hardly "rubbish".

What you describe in 1. is very difficult in practice to execute. Unless you are doing something of great value it is basically a thought exercise.

The amount of time wasted during development due to certificate errors is non-trivial and the ability to ensure that SSL works is pretty important. Most important though is that we don't waste our time on tasks that are unimportant at a given time e.g. having a DNS server, CA and well configured network just to test a simple SSL based service on localhost.

If what you say is true, then _everyone_ should make sure that all of the SSH keys they use are issued by a CA too! I mean, we wouldn't want to make any irresponsible and insecure connections to our servers!

pornel:
MITM attacks are not very difficult, e.g. can be done on public Wi-Fi (or by spoofing a public Wi-Fi hotspot your computer has remembered).

SSH does not behave like `rejectUnauthorized:false`. It verifies identity after first connection (and on the first connection it IS insecure if you manually don't verify the server identity).

You've mentioned localhost. This is an exception — I've asked about it in https://github.com/visionmedia/superagent/pull/832#issuecomment-169758461 but you haven't answered my questions, and argued about PKI/MITM in general.

lgreenlee:
Good Sir, 

Let us reflect on what has just transpired. I provided my support for a feature request but instead received an (unwanted) lecture on MITM attacks and an accusation that I might act "irresponsibly" by using an option that is available in many Linux utilities and applications that deal with SSL based connections. Perhaps openssl, openssh, git, curl svn, firefox and chrome are all "irresponsible" as well for allowing their users to actually get what they need done.

I tried to take the time to explain that there are scenarios where this is OK, however your previous comment would indicate that you did not completely grok my comments. I appreciate that you'd like to better understand my use case, but it is frankly none of your business.

I just need the feature added. Apparently others do as well since they have taken the time to prepare PR's. These PR's, most especially the .tls implementation is the most flexible and would be just what the doctor ordered.

We are your users - your work is popular because you have users but being belligerent and treating them as if they have no clue what they are doing is a good way to alienate them.

pornel:
Your concern may be "I need this, and here's the shortest path to get my stuff done", but my concern is that I'll have to maintain that code. I'm worried that exposing *all* agent options will allow somebody to use options we didn't intend to support, and we don't have tests for. If we break someone's app, we'll be blamed for it.

Proposed names of options for disabling certificate validation IMHO don't make it clear enough that they allow the traffic to be decrypted and modified by attackers. I do think that having a poorly named option that *disables TLS security for all connections to all hosts, even in production*, is irresponsible. Even if you know not to use it outside localhost, others may not and think that some "non-strct SSL" is still secure enough to go through public networks. 

I need to understand motivations behind this to decide whether that's a quick hack or feature I want to support, and see whether maybe we can come up with an alternative better solution that still solves your problem.

Currently I think I'd be OK with having an option that disables SSL security for localhost only (or hosts on local network, but that's a bit trickier to implement), and/or an option with a scary name that disables certificate validation only if `NODE_ENV=development`, so that if it's accidentally left in it won't make real connections vulnerable.

wubzz:
Got caught offguard on this after first having rewritten ~900 tests to use `superagent` instead of `restler`. Honestly baffled how such a simple configuration is being blocked due to a fear of being blamed of something that would obviously be the endusers fault.

Without taking your argument out of context, I would say it is the same as saying a database client library should not allow `DELETE` queries because the lib collaborators could be blamed for accidental `DELETE` queries.

I don't mean to sound disrespectful or to necro this old issue, just need to get my dissapointment out there... `git reset --hard` we go!

Sakaza:
To use the tank analogy mentioned earlier ...

The rejection of this pull request is like manufacturing tanks without hatches, due to the hatch being seen as a security risk. Sometimes, someone needs to get in or out of the tank, man!

MieszkoGulinski:
Would be useful for me too - for requests in internal network not having control over other servers' certificates.

josketres:
Well I guess I will have to change to another http client library.

=(

stefaneg:
Let me get this right. So there is no way for superagent to communicate with a service using a self-signed certificate? Really?

niftylettuce:
`disableStrictSSL` is a new method added in latest release

mamal72:
@niftylettuce Can you please give me a link about this? Or a link to the implementation part/code?

jmarca:
@mamal72 you can't find it with github search because the feature is called `disableTLSCerts()`, not disableStrictSSL

I found documentation here: https://github.com/visionmedia/superagent/blob/master/docs/index.md#tls-options

Seems to work for my case, and many thanks @niftylettuce for adding this.  Makes testing using connected docker containers feasible.

Issue #807, opened 2015-12-05
Document 'progress' event


ArtskydJ:
The `progress` event should be documented. See #801.

``` js
superagent('POST', /logo)
    .set({ 'Accept': 'application/json'})
    .attach('logo', data)
    .on('progress', function (e)  {
        console.log(e)
        // { loaded: 1234, total: 2468, percent: 50 }
        // not sure about `percent` in the client side
    })
    .end()
```

ArtskydJ:
From [`lib/node/index.js`](https://github.com/visionmedia/superagent/blob/a34e6357b2fc6ffec996600b5d7d51d2e1512501/lib/node/index.js#L1044-L1048):
```js
self.emit('progress', {
  lengthComputable: lengthComputable,
  loaded: loaded,
  total: total
});
```

From [`lib/client.js`](https://github.com/visionmedia/superagent/blob/33f2066802fb1e64dd1b9db08837f6da574c4dae/lib/client.js#L964-L969):
```js
var handleProgress = function(e){
  if (e.total > 0) {
    e.percent = e.loaded / e.total * 100;
  }
  self.emit('progress', e);
};
```

PR #837, opened 2016-01-08
Add 'direction' property to 'progress' events

### LINKED ISSUE #803: Distinguish between upload and download progress events
As mentioned in #516 #771 the progress event doesn't contain information to distinguish between upload progress and response download progress.

ArtskydJ:
Fixes #803. (The docs should be added during #807.)

So the client only emits `'progress'` during a download, and the node version only emits `'progress'` during an upload?
I was under the impression that the node version would emit `'progress'` during uploads and downloads, but it doesn't look like it does. And the changelog seems to back that up.

I wonder if `'upload'` and `'download'` are the wrong terms, because they imply where the request/response is going, but what we actually know is whether it is a request or a response.

### DIFF
--- a/test/node/multipart.js
+++ b/test/node/multipart.js
@@ -148,12 +148,14 @@ describe('Request', function(){
     it('should fire progress event', function(done){
       var loaded = 0;
       var total = 0;
+      var direction = '';
       request
       .post(':3005/echo')
       .attach('document', 'test/node/fixtures/user.html')
       .on('progress', function (event) {
         total = event.total;
         loaded = event.loaded;
+        direction = event.direction;
       })
       .end(function(err, res){
         if (err) return done(err);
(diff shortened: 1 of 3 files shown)


test/node/multipart.js
```diff
@@ -148,12 +148,14 @@ describe('Request', function(){
...
       var loaded = 0;
       var total = 0;
+      var direction = '';
       request
       .post(':3005/echo')
       .attach('document', 'test/node/fixtures/user.html')
       .on('progress', function (event) {
         total = event.total;
         loaded = event.loaded;
+        direction = event.direction;
```
pornel: if that endpoint echoes, then the file is downloaded as well. It's a bit worrying that there's no download event after the upload.

ArtskydJ: It looks like node only emits `'progress'` on uploads. Perhaps the `'progress'` event should be added to node downloads as well.

pornel: Can you change the test to be forwards-compatible, e.g set a flag that a desired event has been seen?

ArtskydJ: Done.

pornel:
Thanks

Issue #830, opened 2016-01-05
API consistency and code sharing


focusaurus:
There are numerous open issues noting discrepancies in either API or implementation between the browser and node codebases. I've done a comprehensive review of the public API (at least for methods if not all events). Posting here for review and discussion. Should we extract a module (commonjs file, not a separate npm package) of shared code? If so, how would the documentation work? Should we refactor code to be shared when it's 98% similar already and technically possible to do so?

- [x] `request(method, url)`: duplicated, refactor
- [x] `req.timeout(ms)`: duplicated, refactor
- [x] `req.clearTimeout(ms)`: duplicated, refactor
- [x] `req.use(fn)`: duplicated, refactor
- [x] `req.then(fn)`: duplicated. refactor
- [x] `req.parse(fn)`: duplicated, refactor
- [x] `req.set(name, value)`: slight difference, refactor to shared if possible
- [x] `req.getHeader(name)`: slight difference, refactor to shared if possible
- [x] `req.unset(name, value)`: slight difference, refactor to shared if possible
- [x] `req.field(name, value)`: slight difference, refactor to shared if possible
- [x] `req.abort()`: slight difference, extract base function and extend in subclasses
- [x] `req.withCredentials()`: browser-only API. Node could be no-op stub I think
- [x] `req.send(body)`: slight difference, refactor to shared if possible
- [x] `req.toJSON()`: node only. Refactor to shared?
- [ ] `req.query(value)`: slight difference, refactor to shared if possible
- [ ] `progress event`: see issues #807 and #803
- [ ] `req.get(url, data, fn)` and other HTTP methods: lots of duplication here. Refactor
  - node side supports a lot more HTTP verbs from the "methods" npm package
- [ ] `req.type(value)`: slight difference, refactor to shared if possible
- [ ] `req.accept(value)`: slight difference, refactor to shared if possible
  - for both `.type()` and `.accept()` the node side uses `mime` for dubious benefit
  - currently debating removing mime dependency and keeping it to the 6 aliases the client code supports, OR just leaving the code as is

The following code has been reviewed and is OK as is:

- `req.auth(user, password)`: different implementations. Slight API difference. OK.
- `request.Request(method, url)`: same API, different code, OK
- `req.attach(field, file, filename)`: same API, different code, OK
- `req.end(fn)`: same API, different code, OK
- `req.pipe(stream)`: node only, OK
- `req.write(chunk)`: node only, OK
- `req.redirects(number)`: node only, OK
[...]

PR #839, opened 2016-01-08
sort query string

vicanso:
sort query string for HTTP accelerator, such as varnish

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -994,6 +994,10 @@ Request.prototype.end = function(fn){
     this.url += ~this.url.indexOf('?')
       ? '&' + query
       : '?' + query;
+    var index = this.url.indexOf('?');
+    if (~index) {
+      this.url = this.url.substring(0, index + 1) + this.url.substring(index + 1).split('&').sort().join('&');
+    }
   }
 
   // initiate request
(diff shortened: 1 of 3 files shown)


focusaurus:
@vicanso would you be willing to review and unify query string handling comprehensively now that we have the `request-base.js` file for common code. I believe we could have the same code running in node and the browser, but currently both the behavior and implementations are different. It would take some analysis to describe the differences, choose the most correct behavior, code that into `request-base.js`, update tests and docs. That would be part of #830.

My initial thought was the query string should respect the order values are added across several `.query()` calls, but if there are real caching benefits to be had by sorting, I'd definitely consider it. Generally more deterministic behavior is desirable but I couldn't say with confidence whether there would be negative consequences of sorting by default.

pornel:
I think it's useful to sort query string, but there are problems with this implementation:

* AFAIK it doesn't give any option to have a custom non-sorted order. If someone has settled on another order already, or has a server sensitive to the order, it may be problematic.
* It relies on browser sorting being stable. There's a convention of setting `?foo[]=1&foo[]=2` to represent an array, and order of these identical keys has to be preserved.

I suggest limiting sort by default to a case of using JSON to set the qs, on the basis that order of keys in JSON is not supposed to matter, i.e. sort on `.query({foo:1,bar:2})`, but not `.query("foo=1&bar=2")`.

pornel:
Closing in favor of sortQuery PR

PR #841, opened 2016-01-08
Update dev deps

pornel:
Relaxed version requirements for dev deps, since it can't break production and we have CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,18 +35,18 @@
     "readable-stream": "1.0.27-1"
   },
   "devDependencies": {
-    "Base64": "~0.3.0",
-    "basic-auth-connect": "~1.0.0",
-    "better-assert": "~1.0.1",
-    "body-parser": "~1.9.2",
-    "browserify": "~6.3.2",
-    "cookie-parser": "~1.3.3",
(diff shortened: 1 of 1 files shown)


focusaurus:
These are failing due to saucelabs not being able to start up one of the environments:

```
Error: [init({"build":"724","name":"superagent","tags":
[],"browserName":"iphone","version":"9.2","platform":"Mac 10.10","appium-version":"1.3.6"})] The 
environment you requested was unavailable.: {"status": 13, "sessionId": 
"7421609f14fa4e0c9fcc88e67c6383e2", "value": {"message": "The Sauce VMs failed to start the 
browser or device\nFor more info, please check https://docs.saucelabs.com/reference/troubleshooting-
common-error-messages"}}
```

Can we disable that environment or somehow get a more reliable process?

pornel:
It might be a flaky VM or an actual incompatibility caught by the test. I'm going to restart the test later to see.

focusaurus:
The way I read the error message seems to mean "we couldn't get an environment started at all" thus I don't think they even ran any code in this case.

pornel:
It has worked with Zuul 3.

PR #844, opened 2016-01-11
Default filename to files name

### LINKED ISSUE #843: File name is undefined if not explicitly set.
Using version `1.6.1` when I do:

```javascript
req.attach('image', file); // file is a File object: https://developer.mozilla.org/en-US/docs/Web/API/File
```

The filename becomes `undefined` and the file path becomes `/path/to/filename/undefined`.

However explicitly setting the filename fixes it:

```javascript
req.attach('image', file, file.name);
```

jisaacks:
This fixes #843 

I didn't add tests because I was unsure how to get them running. I got some error about no saucelabs credentials when I tried to run the tests.

__client__
I am assuming the file param is always expected to be a `File` object.

__node__
I am assuming the file param is expected to be an `fs.ReadStream` if not a string.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -791,7 +791,7 @@ Request.prototype.field = function(name, val){
 
 Request.prototype.attach = function(field, file, filename){
   if (!this._formData) this._formData = new root.FormData();
-  this._formData.append(field, file, filename);
+  this._formData.append(field, file, filename || file.name);
   return this;
 };
 
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -197,6 +197,8 @@ Request.prototype.attach = function(field, file, filename){
     if (!filename) filename = file;
     debug('creating `fs.ReadStream` instance for file: %s', file);
     file = fs.createReadStream(file);
+  } else if(!filename && file.path) {
+    filename = file.path;
   }
   this._formData.append(field, file, { filename: filename });
   return this;
--- a/test/client/request.js
+++ b/test/client/request.js
@@ -46,6 +46,24 @@ it('POST native FormData', function(next){
     });
 });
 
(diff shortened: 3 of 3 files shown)


pornel:
You can run node tests with `make test` and browser tests locally with `make test-browser-local`

jisaacks:
> browser tests locally with make test-browser-local

@pornel that worked thanks.

For the tests, I needed a way to check that the name was being sent correctly in the request. The only thing I was able to get back though was a content disposition string.

I tried parsing that string with [this library](https://github.com/jshttp/content-disposition) but was unsuccessful.

I was able to get it to work with a simple regex though. previously the string would have contained `filename="undefined"` and now it contains `filename="image.jpg"` (the actual file name.) I don't have a ton of time to throw at this though but I think this is Good Enough™.

Please let me know.

jisaacks:
Hmmm IE11 seems to be failing at [this line](https://github.com/visionmedia/superagent/pull/844/files#diff-6414ebbe93b17601dcf0eedc1bcc7779R55) in the test.

jisaacks:
OK So it seems [IE11 doe snot support the File constructor](http://caniuse.com/#feat=fileapi) which would explain the failure on that line.

I am not aware of a good workaround. Any ideas?

pornel:
@jisaacks if IE doesn't support it, it's IE's problem. You can skip the test there (see FormData check few lines above)

jisaacks:
@pornel cool. I skipped the test if `File` cannot be used as a constructor. Note, I could not just check for the existence of `File` as in some cases it will exist but not be allowed to be used as a constructor resulting in an error of:

> Error: FileConstructor is not a constructor 

Let me know if there is anything else.

pornel:
Thank you

PR #846, opened 2016-01-12
Add .dataFilter function prior to parsing of body

tquetano-r7:
This function allows you to transform the data prior to parsing of the body. It attempts not to be opinionated on the type of data or what you are trying to do to it. The basic application:
```
request
  .get('url.com/foo')
  .dataFilter((data) => {
    return data.replace(/text_to_strip/g, '');
  })
  .end((err, res) => {
    console.log(res.body); // will have text_to_strip removed globally from the data
  });
```
The original application I was striving for was stripping of XSS prefixes / postfixes, however I tried to make it more generalized so that it could be used for a multitude of purposes. This method is only applied in the parseBody function, and if nothing is returned from the callback passed then it falls back to the original str value passed to parseBody.

Apologies in advance for the lack of commit squashing.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -328,6 +328,7 @@ function params(str){
 
 function Response(req, options) {
   options = options || {};
+
   this.req = req;
   this.xhr = this.req.xhr;
   // responseText is accessible only if responseType is '' or 'text' and on older browsers
@@ -343,7 +344,7 @@ function Response(req, options) {
   this.header['content-type'] = this.xhr.getResponseHeader('content-type');
   this.setHeaderProperties(this.header);
   this.body = this.req.method != 'HEAD'
-    ? this.parseBody(this.text ? this.text : this.xhr.response)
+    ? this.parseBody(this.text ? this.text : this.xhr.response, options.dataFilter)
     : null;
 }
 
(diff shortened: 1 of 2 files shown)


tquetano-r7:
Apologies again, I have no easy way to produce a test for this locally as I don't have saucelabs credentials and therefore cannot do the BROWSER=1 test. Based on another client test, I modified to see if this would work. I promise I have tested it in the real world ...

pornel:
You can run tests without saucelabs with `make test-browser-local`

tquetano-r7:
@pornel Thanks, that is some badassery. I ran that locally and all my tests passed, and I know my travis build failed but I don't think it was due to my test.

pornel:
Yeah, travis tests are sometimes a bit flaky and need restarting.

tquetano-r7:
Ah sweet, noticed the browser test was kicked off again and passed, I assume that was you. :)

focusaurus:
I would prefer not to add more API surface for this. Is this not reasonably achievable already with `request.parse['application/json'] = stripPrefixes`?

In either case this would need consistency in the node/client code before merging.

tquetano-r7:
Going about it in the way you recommended appears to solve my issue. Thanks!

Issue #862, opened 2016-01-20
'del' method bug


Tarlord:
Hello Guys.

Today when we where using  superagent with express together, we found an issue with 'del' method.

Here, you are pushing 'del' method inside methods package, because it is simple required(package) and array of methods is not copied or something - https://github.com/visionmedia/superagent/blob/master/lib/node/agent.js#L58

The problem -  express also using methods package and because it was used after you added 'del' method it comes into an error from express that mentioning that they does not have function to handle 'del' method. 

As i understood how node packages works, you are modifiying 'methods' package directly from yours correct? Is it possible to fix it?(Just make copy of methods list or something). Thanks!

P.S. Sorry for my bad English.

cmordue:
We've run into the same issue.

mrsharpoblunto:
I've posted a pull request which fixes the issue - but for now a workaround is to require() superagent before express. The problem occurs in express because the methods list changes between the time the handler functions are created and then enumerated over - if superagent is required before express, the del method is on the list the whole time and there is no mismatch between the list of handler functions created and the list to enumerate over

pornel:
Fixed in 1.7.1

PR #863, opened 2016-01-20
clone http methods module to avoid change

vicanso:
for #862 , since npm is changed, now some public node_modules may be required by different modules, so  changing module may result in an error.


pornel:
Thank you. We've got this fixed now.

PR #865, opened 2016-01-20
Fixed conflict with express (and possibly other libraries) when using npm 3.x

mrsharpoblunto:
This is caused by superagent modifying the list of http methods coming from the 'methods' package. In npm 2.x this isn't usually a problem as superagent uses its own copy of the methods package, however in npm 3.x all packages are deduped, so express, superagent, and other libraries all share the same package. 

The fix is to create a shallow copy before modifying the list for superagents use so that other libraries are not affected.

### DIFF
--- a/lib/node/agent.js
+++ b/lib/node/agent.js
@@ -55,7 +55,13 @@ Agent.prototype.attachCookies = function(req){
 };
 
 // generate HTTP verb methods
-methods.indexOf('del') == -1 && methods.push('del');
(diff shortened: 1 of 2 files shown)


mrsharpoblunto:
Looks like this should fix https://github.com/visionmedia/superagent/issues/862

focusaurus:
LGTM

lancedikson:
Run into this bug. Waiting for a new version ASAP. Thank you :+1:

PR #868, opened 2016-01-22
Fix case-sensitivity in node header introduced by a4ddd6a

ejinotti:
a4ddd6a introduces case-sensitivity in header set(), get(), and unset(), which is inconsistent with Node HTTP module and the HTTP spec itself. It also borks some superagent addons such as [superagent-hawk](https://github.com/CrowdProcess/superagent-hawk/blob/master/index.js#L31).

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -125,7 +125,7 @@ function Request(method, url) {
   this.method = method;
   this.url = url;
   this.header = {
-    'User-Agent': 'node-superagent/' + pkg.version
+    'user-agent': 'node-superagent/' + pkg.version
   };
   this.writable = true;
   this._redirects = 0;
@@ -249,6 +249,7 @@ Request.prototype.agent = function(agent){
 
 /**
  * Set header `field` to `val`, or multiple fields with one object.
+ * Case-insensitive.
  *
  * Examples:
  *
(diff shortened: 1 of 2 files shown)


focusaurus:
LGTM. I think the build failing is travis flakiness getting one particular environment running.

fabien:
I've ran into an issue this week and I suspect this PR to be at the cause of it. When running the following code, I consistently get a `401` error from a third-party API that used to work fine:

```
var req = request.get(url);
req.set('Content-Type', 'application/json');
req.set('Accept', 'application/json');
req.set('Authorization', 'Bearer ' + key);
req.send(data);
req.end(function(err, res) {
    console.log(res.body);
});
```

The issue started appearing with `1.7.2` for me, so for now I downgraded to fixed `1.7.1` dependency.

Since I have no control over the remote API, I cannot provide any details regarding the wellformedness of the request or particular issues. The problem seems subtle enough to stay under the radar for most SuperAgent users, so I'm curious if others have ever experienced something like this.

pornel:
@fabien can you check which version exactly breaks it? 1.7.2 or 1.8 beta?

fabien:
@pornel apparently `1.7.2` is the latest version I tried, and it had issues. For now, it appears that `1.8.0-beta.2` works fine though! I will have to check more thoroughly.

PR #869, opened 2016-01-23
bump extend dependency due to missing license information

leipert:
For a project of mine it is necessary, that all dependencies have license information.

Unfortunately superagents `extend` dependency does not contain any license information.

Therefore I bumped this dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "cookiejar": "2.0.6",
     "debug": "2",
     "reduce-component": "1.0.1",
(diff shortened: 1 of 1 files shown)


leipert:
@pornel Even with credentials travis fails :+1:

I ran the exact tests locally with node 5.4 and npm 3.3.12 and they did not fail.

pornel:
Thanks

leipert:
Thanks :+1:

PR #874, opened 2016-01-27
avoid TCP port conflicts in all tests

focusaurus:
When starting express servers in unit tests, always listen on a free
port.

This should solve port conflicts entirely, which were still plaguing me (I use a **lot** of ports, especially in 3xxx 4xxx ranges!).

It should also avoid theoretical race condition where server.listen()
takes a long time due to how we use `before`, although in practice
I don't believe this was ever occurring.

There's some boilerplate to do it right with mocha before, etc, but
it's OK.

One implicit interdependency between distinct tests was refactored.

### DIFF
--- a/test/node/agency.js
+++ b/test/node/agency.js
@@ -51,7 +51,14 @@ app.get('/simple', function(req, res) {
   res.status(200).send('simple');
 });
 
-app.listen(4000);
+var base = 'http://localhost'
+var server;
+before(function listen(done) {
+  server = app.listen(0, function listening() {
+    base += ':' + server.address().port;
+    done();
+  });
+});
 
 describe('request', function() {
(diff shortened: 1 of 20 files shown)


pornel:
Great! Maybe this will help with the travis flakyness.

focusaurus:
From the few times I've looked into travis flakiness details, it seems to be travis themselves cannot allocate a Win/IE10 (IIRC) environment for the browser tests, so they never even run. I'm not aware of flakiness with the node tests. But there could be other issues I'm not aware of.

The funny bit was that we had zuul running on port 4000 and some node tests also trying to start a temporary express server on 4000 so if you ran zuul to run the browser tests then tried to run the node tests, they'd always fail, which is pretty facepalmy.

Issue #675, opened 2015-06-10
Fails to parse JSON, doesn't give us the raw response.


ponelat:
For particular reasons, the response from the server is a malformed JSON string, ie: it will fail to parse.
However we'd still like to show the raw body content but can't because superagent doesn't include that data.

Can we have the err callback included the raw response.

Round about here I think...
https://github.com/visionmedia/superagent/blob/master/lib/client.js#L466-L473

black-perl:
Yeah, I am having the same problem. Actually superagent checks the `Content-Type` of the returned `response`. If the content is `text` itself but if the content type is wrong like `application/json` , it tries to parse it. If an error occurs during parsing, `double callback error` is thrown and `res` is not set.

catc:
Same issue here. Response from API is plain text but the header is `text/json` which causes superagent to throw the `double callback` error.

black-perl:
There should be some option to get the raw response text in case the JSON parsing fails ?

reinaldo13:
I added a `.rawResponse` property to the `err` object if the parsing fails. The code is pending approval in pull request https://github.com/visionmedia/superagent/pull/716

black-perl:
@reinaldo13  Great !

FoghostCn:
I had the same problem, content-type is application/json but failed to parse,then it warned double callback!

antwan1986:
Same here; I'm purposefully sending back malformed JSON (as that's what numerous Google scripts do to stop people injecting the resource directly onto a apage in a `script` tag). 

A way to either disable the auto-parsing, or, somehow get the response in the error would allow me to skirt around this problem.

pornel:
#716 added `rawResponse` to the error. Does that solve your problem?

antwan1986:
It'll certainly give a workaround for examining the error and seeing if it "really" is an error. Having some way to disable the auto-parsing sounds like it'd be pretty useful but perhaps the use cases for this functionality are too small.

Do you know when this PR will be included into a release or SuperAgent that could be pulled over NPM? When's your next release?

pornel:
I'll be released soon

titanism:
Malformed JSON is supported: the original body is exposed as error.rawResponse, including browser paths, and regression tests cover it. This was addressed by #716.

v10.4.0 is available on npm: https://www.npmjs.com/package/superagent/v/10.4.0
GitHub release: https://github.com/forwardemail/superagent/releases/tag/v10.4.0

[...]

PR #875, opened 2016-01-28
Wrapped parse call into try/catch block

yunda:
This commit resolves situation when server response holds something like "Forbidden".

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -394,9 +394,18 @@ Response.prototype.setHeaderProperties = function(header){
 
 Response.prototype.parseBody = function(str){
   var parse = request.parse[this.type];
-  return parse && str && (str.length || str instanceof Object)
-    ? parse(str)
-    : null;
+  var result;
+
+  if (parse && str && (str.length || str instanceof Object)) {
+    try{
+      result = parse(str);
+    } catch(e) {
+      result = str;
+    }
+  } else {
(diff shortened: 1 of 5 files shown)


focusaurus:
Could you add a unit test, please? Otherwise seems reasonable to me.

yunda:
This pull-request supposed to fix #675

pornel:
This is a breaking change

pornel:
Returning of unparsed string is dangerous, e.g. let's say the server used to return `{accessDenied: false}` or `Forbidden!`

```js
.end(err, res) {
   if (err || res.body.accessDenied) return;
   allowAccess();
}
```

with this change `res.body` would be "Forbidden!" and `res.body.accessDenied` would be `undefined` and evaluate to falsy value, causing a security issue.

yunda:
@pornel The idea is to return a response even if it fails to parse body. Otherwise we cannot read status codes or any other necessary information form the response object.

pornel:
@yunda you can from the error object, that's what rawResponse was for.

pornel:
@yunda now there's `err.statusCode`. Do you need to read other properties during an error?

pornel:
I'm closing in favor of `err.rawResponse` and `err.statusCode`, which I think are safer and more backwards-compatible.

Issue #876, opened 2016-01-29
Fails to parse JSON, doesn't give us the statusCode


mattdell:
I am attempting to interact with an API that returns 403, a content-type header of `application/json`, but a response body of `Forbidden`.

Because the error response `Forbidden` cannot be parsed it falls into the following case:

```
try {
  res = new Response(self);
} catch(e) {
  err = new Error('Parser is unable to parse the response');
  err.parse = true;
  err.original = e;
  // issue #675: return the raw response if the response parsing fails
  err.rawResponse = self.xhr && self.xhr.responseText ? self.xhr.responseText : null;
  return self.callback(err);
}
```

Since we have the xhr object here, could we add the statusCode as well? I need to handle logic for either a 403, 404, 422, and 500 regardless of the response body.

pornel:
Should be fixed now

PR #877, opened 2016-01-29
Adding statusCode to error response when JSON response is malformed

mattdell:
In reference to issue #876, these changes will allow the client to access the http status code even when the JSON response body is malformed.


mattdell:
The travis build seems to have passed but timed out on the last step. Could someone with write access give it a nudge?

pornel:
Thank you! 

@focusaurus ?

pornel:
I'm going to rebase it to remove unnecessary commits.

mattdell:
Thanks @pornel, that would be greatly appreciated!

mattdell:
I was curious how to do that @pornel and have figured it out. Another PR raised here with just the single commit: https://github.com/visionmedia/superagent/pull/878

PR #883, opened 2016-02-10
XMLHttpRequest for firefox extensions

gonzazoid:
for firefox extensions - there is no way for import XMLHttpRequest into globals. Components.utils.importGlobalProperties (see https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Language_Bindings/Components.utils.importGlobalProperties) does not work properly.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -62,6 +62,9 @@ request.getXHR = function () {
     try { return new ActiveXObject('Msxml2.XMLHTTP.6.0'); } catch(e) {}
     try { return new ActiveXObject('Msxml2.XMLHTTP.3.0'); } catch(e) {}
     try { return new ActiveXObject('Msxml2.XMLHTTP'); } catch(e) {}
+    try { var { Cc, Ci } = require('chrome');
+          return Cc["@mozilla.org/xmlextras/xmlhttprequest;1"].createInstance(Ci.nsIXMLHttpRequest);
+    } catch(e) {}
   }
   return false;
 };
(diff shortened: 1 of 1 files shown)


focusaurus:
I'm hesitant about this. Is there any way we can properly automate tests for this?

Are you using destructuring syntax there? Is that supported in all of our target node and browser environments?

pornel:
Can Firefox extensions solve that problem by setting global XHR before including superagent?

gonzazoid:
>Can Firefox extensions solve that problem by setting global XHR before including superagent?
I've tried import XMLHttpRequest by using Components.utils.importGlobalProperties but this did't solve problem. I'll prepare gist reproducing error (it will take some time)

gonzazoid:
Hi, guys!
I feel so ashamed!
As I said earlier - `Components.utils.importGlobalProperties` does not work properly and it is not possible to import `XMLHttpRequest` into global scope. This is not true. `importGlobalProperties` works as expected, the problem is that we do not have variable `windows` into backround process in firefox extension. This variable is used when checking and not found. This code solves the problem (maybe it will help someone):
```javascript
if(typeof window === "undefined"){

    var viewFor = require("sdk/view/core").viewFor;
    var window = viewFor(ffrequire("sdk/windows").browserWindows[0]);

    let {Cu} = require("chrome");
    Cu.importGlobalProperties(["XMLHttpRequest"]);
    window.XMLHttpRequest = XMLHttpRequest;
}
// we can call superagent here
```

topic can be closed (I'm so sorry I wasted your time)

pornel:
Thanks for the solution

PR #884, opened 2016-02-10
extract common code into request-base object

focusaurus:
This takes any identical copy/paste code between
client code and node code and puts it in a common
superclass.

This handles the identical/easy cases.

Plan to handle the slightly different cases next.

Part of #830

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -4,7 +4,7 @@
 
 var Emitter = require('emitter');
 var reduce = require('reduce');
-var RequestBase = require('./request-base');
+var requestBase = require('./request-base');
 
 /**
  * Root reference for iframes.
@@ -495,7 +495,6 @@ request.Response = Response;
 
 function Request(method, url) {
   var self = this;
-  RequestBase.call(this);
   this._query = this._query || [];
   this.method = method;
   this.url = url;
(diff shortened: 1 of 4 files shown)


focusaurus:
If the collaborators merge this into master, I'll follow up with a series of smaller PRs for each of the methods that have small discrepancies between browser and node code. Just wanted to get the basic structure in place and all tests passing and agreement from collaborators before tackling those.

lib/client.js
```diff
@@ -531,47 +538,15 @@ function Request(method, url) {
 }
 
 /**
- * Mixin `Emitter`.
+ * Mixin `Emitter` and `RequestBase`.
  */
 
 Emitter(Request.prototype);
```
pornel: Should emitter be called on the base only?

focusaurus: Maybe. The complication is the node class inherits from Stream, which inherits from EventEmitter so that's where the event emitter methods come from on the node side, vs the client side which uses component-emitter.

I was planning to just use `events.EventEmitter` and `util.inherits` and let browserify deal with that in the client, but not sure if that would be cleaner. Any idea if we really need the `"browser: {"emitter": "component-emitter"}` bit from `package.json` anymore? I'm not sure our test coverage regarding emitted events is that thorough though so I'm a bit cautious.

pornel: Oh, I didn't realize the Stream complication.

What confused me was that `RequestBase` is called in constructor, which looks like "single inheritance" pattern to me. I'd expect JS classes to either emulate single inheritance, or patch prototypes at will, but not both.

We can't fully use single inheritance pattern (with full prototype chain, Object.create, etc.), so in that case I'd go all the way the other way and make it clear RequestBase isn't a class, but only a bunch of functions to copy.

So my suggestion is to remove call to RequestBase in constructor. Perhaps even instead of using function.prototype, just assign functions to plain object `module.exports = {method: function(){}}`, so it's clear it's not a base class.

focusaurus: OK added a commit along those lines.

pornel:
Great!

focusaurus:
OK, I'll start working through the longer list in #830 and submit individual PRs for each of those methods.

PR #887, opened 2016-02-16
use native xhr auth on client

pyoner:
Use native xhr auth on client because:

> Chrome supports four authentication schemes: Basic, Digest, NTLM, and Negotiate. 

I did check Digest auth on chrome it works

Why this lib did not use native authentication?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -682,13 +682,19 @@ Request.prototype.accept = function(type){
  *
  * @param {String} user
  * @param {String} pass
+ * @param {Boolean} useNative
  * @return {Request} for chaining
  * @api public
  */
 
-Request.prototype.auth = function(user, pass){
-  var str = btoa(user + ':' + pass);
-  this.set('Authorization', 'Basic ' + str);
+Request.prototype.auth = function(user, pass, useNative){
+  if (useNative) {
+    this.username = user;
+    this.password = pass;
+  } else {
+    var str = btoa(user + ':' + pass);
+    this.set('Authorization', 'Basic ' + str);
+  }
   return this;
 };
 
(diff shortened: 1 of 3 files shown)


pornel:
The tests are failing. You can run them locally with `make test-browser-local`

pyoner:
If we use native xhr a chrome browser send username and passwor in url (http://username:password@localhost/) not in HTTP header
I don't know how in Express js get username and password from url to fix test

pyoner:
Added basic auth to server side test

pornel:
I'm little worried that you had to change a test to make it work. What was missing before? Do I understand correctly that now the server has to respond with 401 status first to make the browser to send the `authorization` header?

pyoner:
Yes, you correctly understand

pornel:
Supporting other auth schemes is useful, but OTOH:

* it's a performance regression — it's common to have HTTP APIs that just expect the basic auth header to be there, and can be called in one request
*  this is a breaking change — if someone implemented basic auth API lazily (like our tests) just by checking the authorization header, it'll stop working

Could a hybrid approach work? e.g. if you set Basic auth header and give username/pass to the browser, will the browser switch to a more advanced scheme if needed?

pyoner:
I have better solution:

`req.auth(user, pass, options)`

where **options** is `true` or `object`
if no **options** the library use basic auth (a default behavior)
if **options** is `true` then the library use native auth

if **options** is `object` then look at `options.type`
`options.type` maybe `basic` or `auto` or `digest` (or other auth type need implement)
`auto` means native auth by browser (but I don't know how to implement this on node side, may be we can copy behavior from the browser)

What do you think about it?

pornel:
Yes, sounds like a good idea. Let's go for just `{type:'auto'}` and `{type:'basic'}`, as booleans make calls harder to read.

pyoner:
Done!
Please check!

pornel:
:+1:

PR #888, opened 2016-02-16
added support for blob and arraybuffer responseType

iambumblehead:
Allows the following request to be made
```javascript
request
  .get('http://d8d913s460fub.cloudfront.net/videoserver/cat-test-video-320x240.mp4')
  .set('Content-Type', 'blob')
  .end(function (err, res) {
    console.log(err, res);
  });
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -578,6 +578,26 @@ Request.prototype.type = function(type){
 };
 
 /**
+ * Set responseType to `val`. Presently valid responseTypes are 'blob' and 
+ * 'arraybuffer'.
+ *
+ * Examples:
+ *
+ *      req.get('/')
+ *        .responseType('blob')
+ *        .end(callback);
+ *
+ * @param {String} val
+ * @return {Request} for chaining
+ * @api public
+ */
+
+Request.prototype.responseType = function(val){
+  this._responseType = val;
+  return this;
+};
+
+/**
  * Set Accept to `type`, mapping values from `request.types`.
  *
  * Examples:
@@ -881,6 +901,10 @@ Request.prototype.end = function(fn){
   for (var field in this.header) {
     if (null == this.header[field]) continue;
     xhr.setRequestHeader(field, this.header[field]);
+  }
+
+  if (this._responseType) {
+    xhr.responseType = this._responseType;
   }
 
   // send stuff
--- a/test/client/request.js
+++ b/test/client/request.js
@@ -107,6 +107,16 @@ it('GET querystring object .get(uri, obj, fn)', function(next){
   });
 });
 
+it('GET blob object', function(next){
+  request
+    .get('/blob', { foo: 'bar'})
+    .responseType('blob')
+    .end(function(err, res){
+      assert.deepEqual(res.xhr.response instanceof Blob, true);
+      next();
+    });
+});
+
 window.btoa = window.btoa || null;
 it('basic auth', function(next){
   window.btoa = window.btoa || require('Base64').btoa;


pornel:
Thank you for the pull request. That's a useful functionality.

However, the API for this is a hack. The `.set` method is for HTTP request headers, but this isn't a header, and isn't a property of the request.

We try to keep browser and nodejs API as close as possible, and some users have asked for `Buffer` support as well. Perhaps there could be a single API for enabling `Buffer`/`ArrayBuffer`?

iambumblehead:
@pornel Are you suggesting blob and buffer specific `set` methods? I'll push a different update to this branch which does not use the `set('Content-Type', 'blob')`.

pornel:
Yes, definitely a different method has to be used. I don't have an idea for a good method name for it right now. Maybe someone else can suggest?

iambumblehead:
for reference here is [mdn arraybuffer and blob][0] --i think these are the only two which need to be handled.

what would think of this `set('responseType', 'blob') // blob|arraybuffer`?

[0]: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data

iambumblehead:
It looks like there is a conflict that must be resolved but with the changes I made in the last few minutes, I'm able to make a successful request with this
```javascript
request
  .get('http://d8d913s460fub.cloudfront.net/videoserver/cat-test-video-320x240.mp4')
  .setResponseType('blob')
  .end(function (err, res) {
    console.log(err, res);
  });
```

iambumblehead:
I have run `$ make test-browser-local` as well as `$ make test` --I have 0 failing tests locally and do not understand why the CI build fails.

lib/client.js
```diff
@@ -902,9 +922,17 @@ Request.prototype.end = function(fn){
   // set header fields
   for (var field in this.header) {
     if (null == this.header[field]) continue;
+    // blob and arraybuffer requests use responseType, not header Content-Type
+    // https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data    
+    if (field === 'Content-Type' && this.responseType) continue;
```
focusaurus: I'm definitely not familiar with this aspect of XMLHttpRequest, so I could be way off base here, but this line seems dubious. The `Content-Type` header is describing the body format for the outgoing request and the responseType is indicating how the response body is to be represented to the application JS code. Can you explain how/why `Content-Type` and `responseType` need to interact like this?

focusaurus:
I'd vote for the API `req.responseType(string)` and the implementation just later does `xhr.responseType = whatever` without any  explicit validation of the string value (which might make new valid values "just work" in the future, and allow the error handling to be left up to the browsers). I think exactly matching the official property name `responseType` from the spec is the most obvious thing, just made into a chainable method call because that's the overall pattern of superagent.

If later we want to design some single API for managing binary bodies that is consistent across node and browsers, we could attempt to do that, but it'd be a nontrivial undertaking and we'd need a complete table of all the combinations and design an expressive API. If anyone is anxious to attempt to design that near-term, I'd be OK to postpone this new API until that's ready, but I suspect that pragmatically we should merge this and create a new browser-only API (perhaps with a no-op in node) to make this work for the browser near-term.

lib/client.js
iambumblehead: Thanks for pointing this out. This condition exists somewhat inadvertently as it prevents cycle-http-driver (which I am using) from using a default "json" `Content-Type` when responseType is defined. I'll will remove the condition before committing again.

https://github.com/cyclejs/cycle-http-driver/blob/master/src/http-driver.js#L16

iambumblehead:
@focusaurus per your request `setResponseType` is changed to `responseType`

pornel:
Setting this.responseType, which is same as the method name, will break. Please change it to `_responseType`. Please also add a test.

iambumblehead:
It looks like there may be a problem with the way I rebased. lib/node/index.js is not a file I modified and I do not see the change which appears here in the master branch.

pornel:
Yeah, that doesn't look like a successful rebase. Try `git rebase -i origin/master` and squash/fixup unnecessary commits.

Alternatively pull and checkout master, `git merge --squash` your branch, and make PR from that instead.

iambumblehead:
finally!

pornel:
Great! Thank you @iambumblehead

rohanprasad:
@pornel @iambumblehead Is this issue resolved? If yes, in which version can I use responseType?

pornel:
It's resolved in dbc5d6f8590056c745e0ad9671c7ed6878362e60

Please always use the latest version (currently 2.2)

PR #889, opened 2016-02-17
extract field method to request-base

focusaurus:
deal with root/globals
document both browser and node "val" param types

part of #830

### DIFF
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -2,18 +2,7 @@
  * Module of mixed-in functions shared between node and client code
  */
 
-/**
- * Root reference for environment-specific globals.
- */
-
-var root;
-if (typeof window !== 'undefined') { // Browser window
-  root = window;
-} else if (typeof self !== 'undefined') { // Web Worker
-  root = self;
-} else { // Other environments (node or other)
-  root = {FormData: require('form-data')};
-}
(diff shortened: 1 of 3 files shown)


lib/request-base.js
```diff
@@ -3,6 +3,19 @@
...
+ * Root reference for environment-specific globals.
+ */
+
+var root;
+if (typeof window !== 'undefined') { // Browser window
+  root = window;
+} else if (typeof self !== 'undefined') { // Web Worker
+  root = self;
+} else { // Other environments (node or other)
+  root = {FormData: require('form-data')};
```
pornel: How does it behave when compiled for the browser? Will it cause an unused module to be bundled?

focusaurus: Good point. I checked and it looks like form-data is properly browserified to use the built-in `FormData` on the browser:

https://github.com/form-data/form-data/blob/master/lib/browser.js (their package.json file has the proper "browser" property as well).

I'm OK with this approach overall in dealing with node/browser discrepancies. We'll have a few others to address soon enough like btoa, Buffer, etc.

focusaurus: And actually, based on this it's probably OK to just do a direct `var FormData = require('form-data');` and let browserify do it's thing.

pornel:
neat

PR #891, opened 2016-02-17
header management API consistency

### LINKED ISSUE #873: Inconsistent get-header method for Request in Node and browser
In the browser version's Request function, there's a `getHeader` method to get a request header (that might be set before), and the Node version uses a `get` method for the same purpose.
It would be better to use a consistent name.

focusaurus:
extract common code for get/set/unset/getHeader to request-base.js
extract is-object to standalone commonjs module
fixes #873
part of #830

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -5,6 +5,7 @@
 var Emitter = require('emitter');
 var reduce = require('reduce');
 var requestBase = require('./request-base');
+var isObject = require('./is-object');
 
 /**
  * Root reference for iframes.
@@ -84,18 +85,6 @@ request.getXHR = function () {
 var trim = ''.trim
   ? function(s) { return s.trim(); }
   : function(s) { return s.replace(/(^\s*|\s*$)/g, ''); };
-
-/**
- * Check if `obj` is an object.
- *
- * @param {Object} obj
- * @return {Boolean}
- * @api private
- */
-
-function isObject(obj) {
-  return obj === Object(obj);
-}
 
 /**
  * Serialize the given `obj`.
@@ -498,8 +487,8 @@ function Request(method, url) {
   this._query = this._query || [];
   this.method = method;
   this.url = url;
-  this.header = {};
-  this._header = {};
+  this.header = {}; // preserves header name case
+  this._header = {}; // coerces header names to lowercase
   this.on('end', function(){
     var err = null;
     var res = null;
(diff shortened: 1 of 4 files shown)


focusaurus:
Here's the build failure. I think this is a temporary environment issue. Rerunning.

```
- waiting: <firefox 44 on Mac 10.10>
events.js:141
      throw er; // Unhandled 'error' event
      ^
Error: connection refused: localtunnel.me:42359 (check your firewall settings)
    at Socket.<anonymous> (/home/travis/build/visionmedia/superagent/node_modules/zuul/node_modules/zuul-localtunnel/node_modules/localtunnel/client.js:84:32)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at emitErrorNT (net.js:1256:8)
    at nextTickCallbackWith2Args (node.js:441:9)
    at process._tickCallback (node.js:355:17)
make[1]: *** [test-browser] Error 1
make[1]: Leaving directory `/home/travis/build/visionmedia/superagent'
make: *** [test] Error 2
npm ERR! Test failed.  See above for more details.
```

pornel:
Good stuff!

focusaurus:
Some notes I took while planning/coding this FYI for anyone who may do archaeology on this later:

the header situation

- client
  - this._header stores lowercase names and values
  - this.header stores mixed case names and values
  - .set(name, value) sets both
  - .unset(name) deletes both
  - .getHeader() gets from _header (lower, private API)
  - .end() sets them into the real XHR from .header (preserves case)
  - no req.get() currently, although there is a res.get() on the response
- node
  - this.header stores lowercase names and values (except User-Agent)
  - .request() sets them on the native OutgoingMessage from this.header (uses lowercase names)
  - no .getHeader(), just directly this.header['lower-case-name'] for internal api
  - .get(field) public api is case-insensitve
  - .redirect() deletes them all except 'User-Agent' and sets them again via .set() for second request
- proposed refactorings to unify
  - ✓ this._header stores lowercase names and values
  - ✓ this.header stores mixed case names and values
  - ✓ .set(name, value) sets both
  - ✓ .unset(name) deletes both
  - ✓ .get(name) gets from _header, case-insensitive
  - ✓ .getHeader is an alias for .get, marked as deprecated in next semver major
  - ✓ internal code just does this._header['lower-case-name'] as internal api, don't use .get or .getHeader
  - ✓ internal code when setting headers on native request, use this.header to preserve case of names

pornel:
BTW, I've noticed a bug (that was there before) — `.set('Foo').unset('fOO')` will remove the header from `._header`, but not `.header`.

focusaurus:
Yeah I suspect it's been there a while but no one has encountered/noticed
it. I guess the fix would be to loop through the keys in `.header` and
delete in case-insensitive manner. I can code that up.

Issue #857, opened 2016-01-18
HTTP status 204 with a content-encoding: gzip


vicanso:
HTTP endpoint that returns a 204 with a `content-encoding: gzip`, it will throw Error `[Error: unexpected end of file] errno: -5, code: 'Z_BUF_ERROR'`.

yourcontainer:
+1

focusaurus:
Perhaps we can sanity check `Content-Length` header before unzipping? In node I think we just `.pipe()` into a zlib stream but perhaps that doesn't handle zero-size payload properly. Anyone care to code a unit test and a fix and submit a pull request?

sheltonial:
+1

hustcer:
+

pornel:
I can't fix this without more information. Please don't +1, as this does not help at all.

Please submit unit tests or a full dump of server and client HTTP headers.

hustcer:
@pornel Sorry, I will add more info later.

seangarner:
I doubt this is the cause, but just in case....

A number of conditionals aren't triggered if instantiating the `Request` class directly with a lowercase method.  See visionmedia/supertest/pull/371 for an example of how superagent is doing just this and ending up with the same zlib error as in this issue.

@pornel I'm not sure if you'd consider this a superagent bug or not.  Personally I think it is because the  `Request` class is exported and documented yet methods are only uppercased when calling the function.  If you agree I can open a PR for that.

hustcer:
Just run the following code:
```
const request = require('superagent');

request
    .get('http://www.dfzq.com.cn/dfzq/dfyw/rzrqbdzq.jsp')
    .buffer()
    .end((err, res) => {

        console.log(err, res);
    });
```
and you will get the error:
```
{ Error: unexpected end of file
    at Zlib._handle.onerror (zlib.js:370:17) errno: -5, code: 'Z_BUF_ERROR', response: null } null
```

pornel:
Thanks for the test case with `dfzq.com.cn`. That helped me diagnose and fix the problem.

hustcer:
@pornel Great !

PR #898, opened 2016-02-25
handle HTTP status 204 with content-encoding: gzip

sheltonial:
Fixes issue https://github.com/visionmedia/superagent/issues/857

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -811,8 +811,10 @@ Request.prototype.end = function(fn){
     }
 
     // zlib support
-    if (/^(deflate|gzip)$/.test(res.headers['content-encoding'])) {
-      utils.unzip(req, res);
+    if (res.headers['content-length'] && res.headers['content-length'] > 0) {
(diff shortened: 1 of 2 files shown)


focusaurus:
LGTM. :+1:

sheltonial:
Actually.. I think it needs a little more work. There's `pipe` function which also attempts to unzip.

lib/node/index.js
```diff
@@ -430,7 +430,9 @@ Request.prototype.pipe = function(stream, options){
     }
 
     if (/^(deflate|gzip)$/.test(res.headers['content-encoding'])) {
-      res.pipe(zlib.createUnzip()).pipe(stream, options);
+      if (res.headers['content-length'] && res.headers['content-length'] > 0) {
+        res.pipe(zlib.createUnzip()).pipe(stream, options);
```
pornel: what about `chunked` transfer encoding?

pornel:
Thank you for the PR. I've merged it, but also allowed content-length header to be missing in order to support gzipped chunked encoding.

PR #900, opened 2016-02-25
Add link to superagent-use

koenpunt:
`superagent-use` makes is possible to apply plugins to all requests, instead of just a single request.

My usecase is that I use SuperAgent to request data from an API, which resides on another domain than the application itself. 

With this plugin I no longer have to set the hostname for every request, I simply add a global plugin.

Before:

```js
var superagent = require('superagent');
var prefix = require('superagent-prefix')('https://api.example.com');

superagent
  .get('/')
  .use(prefix)
  .end(function(){});

superagent
  .post('/auth')
  .use(prefix)
  .send({user: 'foo', pass: 'bar'})
  .end(function(){});

// etc..
```

After:

```js
var superagent = require('superagent-use');
var prefix = require('superagent-prefix')('https://api.example.com');
superagent.use(prefix);

superagent
  .get('/')
  .end(function(){});

superagent
  .post('/auth')
  .send({user: 'foo', pass: 'bar'})
  .end(function(){});

// etc..
```


pornel:
Thanks

PR #903, opened 2016-02-27
.abort() extracted to request-base

focusaurus:
- Client code did not return `this` when `.abort()` was called on an
already-aborted instance. This was probably an unreported bug. Now
consistently returns `this` for chaining.
- internal state is called `_aborted` consistently. Any third party
code directly accessing `req.aborted` in the browser will break. This
property was and is an internal property and not a public API.

(heads up @jpodwys)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -535,22 +535,6 @@ for (var key in requestBase) {
 }
 
 /**
- * Abort the request, and clear potential timeout.
- *
- * @return {Request}
- * @api public
- */
-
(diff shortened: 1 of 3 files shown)


jpodwys:
Thanks!

focusaurus:
Trying to confirm with the tests that `.abort()` is chainable but getting some unexpected failures where the `.end` callback is firing in browsers. Give me some time to investigate. May need to revert this.

PR #904, opened 2016-02-28
test for multiple abort calls

focusaurus:
Turns out positioning of `this.clearTimeout()` matters in the browser.

Sorry for the churn on this one. I'm pretty confident it is finalized now.


pornel:
Tests are always good :)

PR #906, opened 2016-02-28
extract .toJSON() to request-base

focusaurus:
This API is now present in both node and the browser.
It was node-only prior to this.
Docs improved and unit tests moved so it runs in node and browser.

Part of #830


pornel:
:+1:

jpodwys:
Thanks for doing this!

PR #909, opened 2016-03-02
Handle modified redirects

pornel:
Avoid inconsistencies when callbacks for the redirect event modify url or headers

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -535,11 +535,11 @@ Request.prototype.redirect = function(res){
 
   // redirect
   this.url = url;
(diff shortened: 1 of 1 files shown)


pornel:
This changes which url's query string is visible in the event, so it may be a break.

focusaurus:
Code change is pretty obviously not problematic, so LGTM. Could you explain a bit further the before and after behavior driving this change?

pornel:
I wanted

    request.get('/this will redirect/').on('redirect', () => req.url = req.url.replace(foo,bar))

and have the correct url in `response.redirects`.

PR #910, opened 2016-03-02
Upgrade the qs package

### LINKED ISSUE #908: qs dependency is REALLY outdated
Current `qs` dependency does not correctly encode `'` in query strings. but current version does. 

After further investigation it would appear that `qs` itself is inconsistent between the browser and node. Output in [the browser](https://tonicdev.com/56d5fffd67c99d0d00e23fbc/56d62322aa56a10d00254857) does not match output in node. 
image

clayreimann:
I didn't address the myriad of other outdated dependencies, but can if that would be useful.

Closes #908.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "url": "git://github.com/visionmedia/superagent.git"
   },
   "dependencies": {
-    "qs": "2.3.3",
+    "qs": "^6.1.0",
     "formidable": "~1.0.14",
     "mime": "1.3.4",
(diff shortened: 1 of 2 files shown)


focusaurus:
LGTM. Thanks!

pornel:
Does this require us to bump major semver?

clayreimann:
It might if you think that no longer returning an error when UTF-16 characters will be encoded in the query string constitutes a breaking change

PR #912, opened 2016-03-02
clean up and comment flaky pipe unit test

focusaurus:
the setTimeout is a hack based on the hypothesis of filesystem sync
race conditions on Travis CI.

Would love to see a PR with true root cause understood and rock-solid
reliability on Travis CI.

Hoping this hack provides temporary relief.

### DIFF
--- a/test/node/pipe-redirect.js
+++ b/test/node/pipe-redirect.js
@@ -31,11 +31,16 @@ before(function listen(done) {
 });
 
 describe('pipe on redirect', function () {
-  afterEach(removeTmpfile);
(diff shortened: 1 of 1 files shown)


pornel:
maybe this *is* an actual bug? Do we have to wait for some kind of end/close event?

focusaurus:
Sounds plausible to me. Maybe we should wait until stream fires "end" before we emit "end" [here](https://github.com/visionmedia/superagent/blob/698e2f2acd76bce0bdac6ce4cd0981a6e9e10963/lib/node/index.js#L364)?

PR #913, opened 2016-03-02
fix pipe race condition

focusaurus:
This should be the more correct event flow to avoid emitting "end"
before the destination stream has been fully written to.

Suspect this will resolve flakyness of TravisCI tests.

See also #912

This passes but I'm not super confident about the correctness of this change. Would appreciate a code review from @defunctzombie or anyone else well-versed in the stream API and superagent.

### DIFF
--- a/test/node/pipe-redirect.js
+++ b/test/node/pipe-redirect.js
@@ -31,30 +31,25 @@ before(function listen(done) {
 });
 
 describe('pipe on redirect', function () {
-  afterEach(removeTmpfile);
+  var destPath = 'test/node/fixtures/pipe.txt'
+  afterEach(function removeTmpfile(done) {
+    fs.unlink(destPath, function (err) {
+      done();
+    });
+  });
   it('should follow Location', function (done) {
-    var stream = fs.createWriteStream('test/node/fixtures/pipe.txt');
+    var stream = fs.createWriteStream(destPath);
     var redirects = [];
     var req = request
       .get(base)
       .on('redirect', function (res) {
         redirects.push(res.headers.location);
       })
-      .on('end', function () {
-        var arr = [];
-        arr.push('/movies');
-        arr.push('/movies/all');
-        arr.push('/movies/all/0');
-        redirects.should.eql(arr);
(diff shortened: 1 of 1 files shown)


focusaurus:
If this gets merged, #912 can be closed without merge.

defunctzombie:
This seems weird to me in the sense that you are waiting for a finished even on the stream you are piping to when usually that stream is what would wait on your end event no?

focusaurus:
I think it's confusing here. I don't think our implementation of `pipe` is really part of the node stream API. It's just sugar for piping the *response** to a file, but you call `req.pipe` and listen for `req` to emit `end`. So yeah, if we had exposed the *res* to the caller and let the caller do `res.pipe(writable)`, the caller could listen for `end` or `finish` etc according to the node streaming API. However, I think the `end` fired off `req` is not really the same thing, so with this change I'm making the semantic "when the response has been fully streamed to its destination stream and that stream is finished, we are clear to indicate via `end` to our callers that the IO is done and they can proceed.

Does that make sense? Again, I could be completely misunderstanding this, but like I said I think it's confusing because there's real streams here and even though `Request` subclasses `Stream` I'm not sure it's going to actually behave like a stream with all the required methods and events. For example, is `req.pause()` going to do anything useful?

focusaurus:
Maybe it's the unit test that is broke then? Maybe the unit test needs to do `stream.on('finish')` before trying to read the file instead of binding reading the file to then `req` emits `end`?

defunctzombie:
Makes sense. In that case tho I would expect that listening to "end" on the response is the right thing to be doing with respect to the response being done (unless that isn't actually indicative of the response being done).

Does changing the test to wait for the end even on `stream` make it work reliably?

focusaurus:
I just rewrote this and rebased as just a fix to the test logic leaving the superagent logic unchanged. We'll see if it resolves the TravisCI reliability problems. I would say this has a fair shot of fixing the issue for good but would definitely welcome anyone else to review the problem and offer their own analysis. @defunctzombie thanks for your input!

focusaurus:
OK my git/github/rebase kung fu failed me repeatedly, so closing #912 and this one in favor of #914

PR #914, opened 2016-03-03
fix race condition in pipe tests

focusaurus:
wait for destination stream to finish writing before reading it back to
verify

Suspect this will resolve flakyness of TravisCI tests.

See also #912 (which I think should now be closed without
merge)

(Sorry, created a new PR instead of #913 due to my git/github rebase-fu failing repeatedly. See #913 for the backstory and #912 will be closed)


focusaurus:
Whew...it actually passed the first time. Had fingers crossed for that one.

PR #918, opened 2016-03-04
Fix wrong jsdoc of Request#attach

Gerhut:


pornel:
Thank you. Docs improvements are always welcome!

PR #919, opened 2016-03-04
add sortQuery function

vicanso:
sort query string for HTTP accelerator, such as varnish, instead of #839

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -6,6 +6,7 @@ var Emitter = require('emitter');
 var reduce = require('reduce');
 var requestBase = require('./request-base');
 var isObject = require('./is-object');
+var isFunction = require('./is-function');
 
 /**
  * Root reference for iframes.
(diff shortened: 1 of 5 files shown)


focusaurus:
Would `sortQuery` be a better name in terms of English grammar verb object?

vicanso:
@focusaurus I have changed `querySort` to `sortQuery `

focusaurus:
LGTM. Semver-Minor. @pornel thoughts?

pornel:
`'function' == typeof this._sort` is overengineered. `.sortQuery(false)` doesn't disable sorting.

focusaurus:
I rebased this code and addressed Pornel's comments. The tests now pass and there's documentation. It's in my fork here: https://github.com/focusaurus/superagent/tree/sort-query

@vicanso The rebase lost your commit author metadata. You may want to take the final diff between my sort-query branch and master, make a single new feature commit, and then PR that in. I think we should be able to accept the PR then unless @pornel has any further reservations.

pornel:
Regarding git, instead of committing, you need to run `git rebase --continue` to preserve authorship. You can fix this with `git commit --amend --author="user <email>"`.

As for the code, I don't see the point of using is-function module instead of `'function' === typeof`. From what I gather it's a workaround for old browsers confusing regexes with functions, none of which applies here.

pornel:
rebased as 07147ce74ccc4a87388593c68b2d8a94d3d0bc41

PR #924, opened 2016-03-07
Rename @api private properties to _prefixed

### LINKED ISSUE #922: Add underscore prefix to names of methods with "@api private"
Currently we have private methods that look like public methods due to having "normal" names and only "@api private" comment. Using `_`-prefix convention will make it clearer.

pornel:
Fixes #922


focusaurus:
LGTM

Issue #854, opened 2016-01-15
Superagent 2.0


pornel:
There are few issues/pull requests with [changes/fixes that could potentially be backwards-incompatible](https://github.com/visionmedia/superagent/labels/Semver-Major).

* Should we make a 2.0 release?
* Should we continue developing/backporting fixes to 1.x? If so, what about 0.x?
* Is it better to release all breaking changes at once in a really major 2.0, or should changes be spread to break little at a time, and be released as v2, v3, v4... v20, v21... etc.?

focusaurus:
I can't weigh in from tons of experience on libraries of this popularity, but my gut says the substack/hammer approach of shipping semver major as often as we need is better. I think browserify has gone from v9 to v13 and I've never had to change my code. Ditto for joi.

Given this library still has a pretty big backlog of PRs and issues and a fairly small allocation of contributor attention, I don't think we're in a position to offer "long term/maintenance" releases on prior major versions. I think for the most part, stuff that's broken in v1 has probably been broken a while.

I'm OK with batching up a few issues into a single semver release if we already know about them and have a fix plan at the same time, but other than that I don't think we should hold code back more than say 2-3 weeks.

olalonde:
Sorry, I'm a bit out of the loop here as I've been busy with other projects but I also vote for releasing major versions more often (as needed) and only consider long term maintenance if it becomes a real issue to library users.

On a related note, I'd like to propose we use babel as part of the build so we can safely start using es2015 features.

focusaurus:
I'm -1 on babel for the node code base. I guess I'm OK with it for the browser, but I'd rather build the client via browserify sticking with ES5 as we currently do. I find needing a babel preprocessor before running code on node a major development hassle and I'd rather say this works on node directly.

olalonde:
I was also initially skeptic at first but I have been using babel for the last few months in other projects. Also babel comes with `babel-node` which is a drop in replacement for `node` when developing. We would just need a prepublish script that runs babel before publishing to npm.

focusaurus:
It's not that simple. You need to transpile before you can run tests or debug. You need sourcemaps. Many existing tools for debugging, linting, code coverage, etc are not babel aware yet and everything becomes slower and more complicated and confusing. All this makes babel painful for me and ES5 simple and fast. I don't think there's much in the ~3000 LoC superagent codebase that would hugely benefit from ES2015 features. Anyway, end of my opinion on that one.

focusaurus:
[...]

PR #925, opened 2016-03-08
Make .then() return a real Promise

### LINKED ISSUE #722: The .then() function is not compatible with Promises

### LINKED ISSUE #852: Fake Promise is a breaking API change
The fake promise that was introduced in #709 is a **BREAKING** change and not backwards compatible.

I wrote a little scenario involving an async action via a promise. Before this change was made, this was perfectly valid until the fake promise was added to the library. For this scenario, I used native promises but have also tested with a promise library with the same result.

```js
var http = require('http')
var fs = require('fs')
var superagent = require('superagent')
var dest = fs.createWriteStream('response.txt')

// stand up a simple http server
var server = http.createServer(function(req, res) {
  console.log('got request')
  res.end('hello superagent!')
})

// Start listening
server.listen(1234, function() {

  console.log('server listening')
  
  // Generate dummy data
  var promise = new Promise(function(resolve, reject) {
    
    resolve({ hello: 'superagent' })
    
  }).then(function(data) {
    
    // generate a dummy request
    return superagent.post('http://localhost:1234').send(data)
    
  }).then(function(res) {
    // At this point, we should have a request
   // that hasn't been fired yet, based on the docs.

    // Pipe to the dest, fires the request and streams to the file
    res.pipe(dest)
    // kill the program
    dest.once('close', function() {
      console.log('wrote response')
      process.exit()
    })
    
  })
  
})
```

With `superagent@<1.3.0`, the code works as expected and you will see a `wrote response` and the `response.txt` will have the correct text `hello superagent!`. 

The problem with `superagent@^1.3.0` the request ( that should not have fired based on the docs ) is seen as a "thenable" and so the promise chain will call the `then()` method which fires off the request and returns the buffered response to the next promise chain call. Because of this, the pipe method will not work because the stream has already been exhausted, the program will never end, and the `response.txt` will have no content.

pornel:
Another take on Promises. 

* Uses global `Promise`. ES6 is final now, so I think it's OK to expect it for interoperability. 
   Node 0.10—the IE6 of Nodes—can easily polyfill it.
* Re-returns the same result, so that usage of the request as thenable in multiple places won't make multiple calls. That's consistent with how real promises behave.

Fixes #722 Closes #852
 
#854 #726

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -64,6 +64,36 @@ describe('request', function(){
         done();
       });
     })
+
+    it('is optional with a promise', function() {
+      if ('undefined' === typeof Promise) {
+        return;
+      }
+
+      return request.get(uri + '/login')
+      .then(function(res) {
+          return res.status;
+      })
+      .then()
+      .then(function(status) {
+          assert.equal(200, status, "Real promises pass results through");
+      });
+    });
+
+    it('called only once with a promise', function() {
+      if ('undefined' === typeof Promise) {
+        return;
+      }
+
+      var req = request.get(uri + '/unique');
+
+      return Promise.all([req, req, req])
+      .then(function(results){
+        results.forEach(function(item){
+          assert.equal(item.body, results[0].body, "It should keep returning the same result after being called once");
+        });
+      });
+    });
   })
 
   describe('res.error', function(){
@@ -81,6 +111,20 @@ describe('request', function(){
         assert(err, 'should have an error for 500');
         assert.equal(err.message, 'Internal Server Error');
         done();
+      });
+    })
+
+    it('with .then() promise', function(){
+      if ('undefined' === typeof Promise) {
+        return;
+      }
+
+      return request
+      .get(uri + '/error')
+      .then(function(){
+        assert.fail();
+      }, function(err){
+        assert.equal(err.message, 'Internal Server Error');
       });
     })
   })
@@ -312,6 +356,10 @@ describe('request', function(){
 
   describe('.then(fulfill, reject)', function() {
     it('should support successful fulfills with .then(fulfill)', function(done) {
+      if ('undefined' === typeof Promise) {
+        return done();
+      }
+
       request
       .post(uri + '/echo')
       .send({ name: 'tobi' })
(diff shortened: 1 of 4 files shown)


test/basic.js
```diff
@@ -64,6 +64,36 @@ describe('request', function(){
         done();
       });
     })
+
+    it('is optional with a promise', function() {
```
focusaurus: I think you need to take a `done()` parameter here or mocha will consider the test passed before all the `assert` calls are reached.

focusaurus:
I don't see `.then()` mentioned anywhere in `docs/index.md`. Let's document this while we're at it. Other than docs and my comment above about `done()` this looks good to me.

test/basic.js
pornel: I'm using Mocha's Promise support here: https://mochajs.org/#working-with-promises

focusaurus: Ah, I haven't seen that before. Looks good then.

ewnd9:
Is it possible to access a response body in the catch method (e.g. server is responding with a 500 status code, but an error reason is in a body)?

pornel:
I many cases we're adding properties to the error object. Check for `err.response`.

PR #932, opened 2016-03-09
Dep semver updates

focusaurus:
Update to current latest version of all deps.

Consistently use semver major number only.

Code updates for removed should.js APIs.

Didn't update `qs` as that is trickier and being dealt with in a separate issue.

### DIFF
--- a/docs/test.html
+++ b/docs/test.html
@@ -338,7 +338,7 @@ setTimeout(function() {
     should.not.exist(err);
     res.should.have.status(200);
     should.not.exist(res.headers['set-cookie']);
-    res.text.should.include('dashboard');
+    res.text.should.containEql('dashboard');
     done();
   });</code></pre></dd>
             <dt>should start with empty session (set cookies)</dt>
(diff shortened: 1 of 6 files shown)


pornel:
This also allows _downgrade_ of existing dependencies where major version hasn't changed. Could you use ^x.y.z syntax for these?

focusaurus:
Done.

When the travis browser test fails with ` failed: <internet explorer 9 on Windows 2008> (0 failed, 0 passed)` is that a travis issue allocating a worker to run that OS/browser?

pornel:
Great.

Yeah, looks like a saucelabs having problems launching IE.

pornel:
I've restarted the build a few times, and IE9 fails consistently. Maybe it's the FormData issue? Or some other dependency? 

Can you rebase on master? And/or try updating only half of the depencies at a time?

focusaurus:
I rebased but still seeing `failed: <internet explorer 9 on Windows 2008> (0 failed, 0 passed)`. That seems like saucelabs isn't even running the code, but I will start bisecting the changes.

focusaurus:
Now the tests are passing in CI. I don't think it's flakiness in the superagent code, I think it was just an operational issue with saucelab spinning up IE9 workers.

PR #934, opened 2016-03-10
Require form-data on-demand, so that it doesn’t throw when not in use

### LINKED ISSUE #933: 1.8.0 relies on a `form-data`, that expects window.FormData to exist. Throws otherwise
The recent bumped version `1.8.0` broke our CI server...
I don't have too much time to go into details, but this test (with a webpack builder) fails with...

details:
Using  `webpack` + `karma` + `jsdom`

`FormData is not defined <- webpack:///~/form-data/lib/browser.js:1:0`

```js
// Code
import request from 'superagent'
import { expect } from 'chai'

describe('superagent is broke', function(){
  it('should work', function(done){
    request.get('google.com').end(function (err, res) {
      expect(res.test).to.match(/google/)
      done()
    })
  })
})
```

Running a similar test works in node.js

I changed to `1.7.2` and all is well.

pornel:
Fixes #933

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,7 @@
+# 1.8.1 (2016-03-14)
+
+  * Fixed form-data incompatibility with IE9
+
 # 1.8.0 (2016-03-09)
 
  * Extracted common code into request-base class (Peter Lyons)
--- a/bower.json
+++ b/bower.json
@@ -1,4 +1,7 @@
 {
   "name": "superagent",
-  "main": "lib/client.js"
+  "main": "lib/client.js",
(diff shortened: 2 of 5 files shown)


focusaurus:
LGTM. Maybe we can add a webpack build to superagent's CI build/test process to catch similar errors in the future?

pornel:
I don't think this is a problem with webpack—it compiled browser version OK this time. It's a problem that form-data assumes `FormData` exists in the browser.

I don't know how our tests passed in IE9 with that though :)

PR #938, opened 2016-03-15
Set https.servername from Host header

kornelski:
I need to make requests to a specific IP address. With HTTP that's easy:

```
request.get('http://1.2.3.4').set('Host', 'example.com')
```

but for HTTPS it currently fails, because TLS will be validated against `1.2.3.4` rather than `example.com`.

This change makes Host header (if set) take precedence over the URL.

Open questions:
- Is it secure?
- How to test it?

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -518,6 +518,11 @@ Request.prototype.request = function(){
   options.ca = this._ca;
   options.agent = this._agent;
 
+  // Allows request.get('https://1.2.3.4/').set('Host', 'example.com')
+  if (this._header['host']) {
+    options.servername = this._header['host'].replace(/:[0-9]+$/,'');
+  }
+
   // initiate request
   var mod = exports.protocols[url.protocol];
 


jotto:
> Is it secure?

Isn't this the purpose of SNI? (multiple hosts, 1 IP) - this just seems like an alternative API to what would could easily be achieved with DNS or merely calling the lower Node APIs manually.

If there's still paranoia, maybe start by limiting to localhost only (testing environments).

Or maybe I'm misinterpreting what you mean by "is it secure?".

> How to test it?

inspiration from https://github.com/visionmedia/superagent/blob/eafc6583dcac73f04754f2aabc8bce6ea65d00ef/test/node/https.js#L30-L58

```javascript
server = https.createServer({
  SNICallback: function(hostname, cb) {
    if (hostname === 'example.com') cb(null, tls.createSecureContext({ cert, key }));
  }
}, app);

before(function listen(done) {
  server.listen(0, function listening() {
    base += ':' + server.address().port;
    done();
  });
});

describe('https', function(){
  describe('request', function(){
    it('should give a good response', function(done){
      request
      .get(base)
      .set('Host', 'example.com')
      .ca(cert)
      .end(function(err, res){
        assert(res.ok);
        assert('Safe and secure!' === res.text);
        done();
      });
    });
  });
});

```

PR #940, opened 2016-03-16
Listen for FormData error events

### LINKED ISSUE #939: Add listener for formData error event
When sending multipart data, if I give a non-existent path or some filesystem error happens in readStream, `request.on('error')` doesn't receive any error. This causes process to exit non-zero.

If anyone could confirm that this should be fixed/improved, I'm going to open a pull request.

scriptype:
Will fix #939.

I've added event handlers for FormData to wherever it's initiated for the first time.

A simple ENOENT test case is also added. With this change, any FileSystem error will be caught, not only ENOENT.

### DIFF
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -164,10 +164,6 @@ exports.field = function(name, val) {
   if (!this._formData) {
     var FormData = require('form-data'); // browserify compatible. May throw if FormData is not supported natively.
     this._formData = new FormData();
-    this._formData.on('error', function(err) {
-      this.emit('error', err);
-      this.abort();
-    }.bind(this));
   }
   this._formData.append(name, val);
   return this;
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -164,6 +164,10 @@ exports.field = function(name, val) {
   if (!this._formData) {
     var FormData = require('form-data'); // browserify compatible. May throw if FormData is not supported natively.
     this._formData = new FormData();
+    this._formData.on('error', function(err) {
+      this.emit('error', err);
+      this.abort();
+    }.bind(this));
   }
   this._formData.append(name, val);
   return this;
--- a/test/node/multipart.js
+++ b/test/node/multipart.js
@@ -177,6 +177,21 @@ describe('Request', function(){
         done();
       })
     })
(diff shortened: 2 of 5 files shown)


lib/request-base.js
```diff
@@ -164,6 +164,10 @@ exports.field = function(name, val) {
   if (!this._formData) {
     var FormData = require('form-data'); // browserify compatible. May throw if FormData is not supported natively.
     this._formData = new FormData();
+    this._formData.on('error', function(err) {
```
pornel: This may not work in the browser, where we use native `FormData` without `on` methods

pornel: Looks like we don't have a browser test that covers this method. Could you add a shared test that calls 

```js
request.post('/upload')
     .field('foo', 'bar')
     .end(callback);
```

and see if that works? (with `make test-browser-local`)

scriptype:
@pornel This is actually completely node-side. Users will probably handle producing buffer themselves or most likely file will be selected through a file input, thus it's all OS's job.

We must make sure calling `on` method will not break browser though.

scriptype:
I've tried adding shared test for field but regardless of my changes, it was failing. And adding listener to `request-base` was breaking the browser. So I've removed the listener from `request-base` and the shared test of `field` method.

Currently it catches error on node, doesn't behave differently on browser.

test/node/multipart.js
```diff
@@ -177,6 +177,20 @@ describe('Request', function(){
         done();
       })
     })
+    it('filesystem errors should be caught', function(done){
+      var timestamp = Date.now()
+      request
+          .post(base + '/echo')
+          .attach('filedata', 'test/node/fixtures/lorem-' + timestamp + '.ext')
```
focusaurus: If this is supposed to be a non-existent file, let's make it simple and obvious and just use `'test/node/fixtures/non-existent-file.ext'`. The timestamp implies something about uniqueness that I don't think is really at play here.

focusaurus:
Other than my one comment about the unit test filenames, this seems like good functionality to merge. Thanks!

test/node/multipart.js
scriptype: Good point. I've thought something similar later on. Let me clear that.

PR #942, opened 2016-03-16
Rename ambiguous .parse to .serialize

### LINKED ISSUE #805: Possible inconsistent method: parse
I've run into some trouble while trying to customize the way response bodies are parsed -- specifically on the client.

From [these tests](https://github.com/visionmedia/superagent/blob/master/test/node/parsers.js#L22-L74), it appears that the `parse` method of the node library is used to customize how *response* bodies are parsed.

[This test](https://github.com/visionmedia/superagent/blob/master/test/client/request.js#L720-L738) shows that the `parse` method of the client library is used to customize how *request* bodies are serialized.

Is this inconsistency intentional?  If not and a maintainer would like to specify the intended behavior, I'd be happy to open a PR.

pornel:
Fixes #805

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -193,6 +193,10 @@ You can also use the `.query()` method for HEAD requests. The following will pro
      request.post('/user')
        .type('png')
 
+## Serializing request body
+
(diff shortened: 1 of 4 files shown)


focusaurus:
LGTM

Issue #779, opened 2015-11-09
UNIX socket path in node.js


enricostano:
Hi, is there a way for `superagent` to make a HTTP request using UNIX sockets instead of `host:port`?

Thanks!

focusaurus:
I believe that may be technically possible but reviewing the current code I'm pretty sure it does not currently work. This would create another feature disparity between the browser and node, so there may be some objection to adding this based on that. My thought would be getting the basic mechanics working over a unix socket would be pretty straightforward but then it would get complex dealing with redirects and facets of HTTP coupled to the host:port origin like cookies etc.

olalonde:
Personally, I wouldn't be opposed to supporting that if it's just a few lines of code.

kanreisa:
@enricostano
Hello! I've implemented it on https://github.com/kanreisa/superagent/commit/9e57664dc130d46fcc17a47301cf8756da339a55 .
If you sure, try it and feedback please.

**Usage:**
```
http://unix:/absolute/path/to/unix.socket:/request/path
```

enricostano:
hi @kanreisa thanks! But I finally decided to create 2 different and separate API clients, one for server side (using `request`) and another for browser side (using XHR, until fetch doesn't get stable enough).

kanreisa:
@enricostano np ;)

PR #961, opened 2016-04-04
Support unix sockets

kanreisa:
related: https://github.com/visionmedia/superagent/issues/779

I've implemented it. maybe this is not breaking change.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -48,9 +48,10 @@ ES6 promises are supported. Instead of `.end()` you can call `.then()`:
  
  The __node__ client supports making requests to [Unix Domain Sockets](http://en.wikipedia.org/wiki/Unix_domain_socket):
 
-     // pattern: http://unix:/SOCKET_PATH:/REQUEST_PATH
+     // pattern: https?+unix://SOCKET_PATH/REQUEST_PATH
+     //          Use `%2F` as `/` in SOCKET_PATH
      request
-       .get('http://unix:/absolute/path/to/unix.sock:/search')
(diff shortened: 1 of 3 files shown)


pornel:
Thanks for the PR, but that is a hack.

I think it's inappropriate to use `http:` URL scheme for something that is not true HTTP over TCP/IP. `unix` is not a special, reserved hostname.

Python uses `http+unix` scheme for this: https://pypi.python.org/pypi/requests-unixsocket/ which seems more appropriate and avoids ambiguity.

kanreisa:
@pornel Thanks review, I've fixed.

pornel:
That's better. Thanks! Can you add tests to ensure we don't break it?

kanreisa:
@pornel yes, added :cake:

philpill:
@kanreisa is this PR the cause of the current build failure?Is anyone looking into it?

kanreisa:
retry build please

pornel:
I've restarted the build

kanreisa:
Oops, Node.js 0.10 has failed. I'll check it later.

philpill:
Sorry, I added the comment about the failing build a bit earlier, but removed it as I'm not sure if this PR (or the work for this PR) was the cause.

pornel:
Have you checked node 0.10? If there's no simple fix for it I don't mind disabling the feature in node 0.10 (i.e. detect node 0.10 in tests and skip the tests).

pornel:
Since 0.10 life [ends this month](https://github.com/nodejs/LTS#lts_schedule) we can go ahead and merge this.

PR #962, opened 2016-04-04
refactor: add _appendQueryString to client request

vkarpov15:
Heya,

Server-side requests have this `_appendQueryString()` function, but client-side requests don't, so there's no consistent way for plugins to apply the query string. Makes things a little annoying for plugins that overwrite `Request.prototype.end()` like https://github.com/vkarpov15/superagent-httpbackend.

Thanks,
Val

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -636,6 +636,9 @@ Request.prototype.auth = function(user, pass, options){
 */
 
 Request.prototype.query = function(val){
+  if (arguments.length === 0) {
+    return this._query.join('&');
+  }
   if ('string' != typeof val) val = serialize(val);
   if (val) this._query.push(val);
   return this;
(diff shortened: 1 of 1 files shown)


pornel:
Thanks. Consistency between node and client is definitely a good idea.

I'm unusure about the sneaky `.query()` getter, especially that it isn't consistent.

vkarpov15:
Ah oops, my mistake re: the query() getter, thought about adding that but decided against it since what I really wanted was a way to consistently apply the query string between client and server.

vkarpov15:
@pornel quick reminder to take a look at this when you get a chance :+1:

philpill:
Is anyone looking into why this broke the build?

pornel:
The code looks OK now, but for some reason the tests are failing. I've restarted the build.

vkarpov15:
Looks like the build is breaking because of some tunneling issue. Travis being flakey?

pornel:
It keeps timing out, and not only due to the tunnelling. Suspicious. I'm going to check if it passes otherwise #967

pornel:
Sorry about the delay. Looks like our tests are broken #972, so I'm merging without the tests (code looks ok)

PR #1002, opened 2016-06-08
Don't determine XHR every time

lmk123:
I think it will improve some performance.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -36,19 +36,21 @@ var request = module.exports = require('./request').bind(null, Request);
  * Determine XHR.
  */
 
-request.getXHR = function () {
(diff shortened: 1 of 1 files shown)


focusaurus:
Isn't this going to fail here: https://github.com/lmk123/superagent/blob/81c979d43f32c5016917475a8e58a1ecfc40ff49/lib/client.js#L755 ? You removed a function without updating the places that call that function.

lmk123:
@focusaurus I'm not removed this function `request.getXHR()`. It is strange that when I run `make test-browser-local`, everything is ok. I will take a look.

lmk123:
I can't figure out why it's failed on Travis CI.

Issue #313, opened 2014-01-21
remove annoying "double callback" warning


stephenmathieson:
i'd rather have an error thrown than `console.warn('double callback!')`

kfng7:
I also got the same issue when the remote server is down. I expect that error message should be passed via callback instead.

lukebayes:
Also experiencing this issue using gulp watch with mocha.

```javascript

var request = require('supertest'),
      worker = require('../src/worker');

describe('POST /message', function() {
  var app;

  beforeEach(function() {
    app = worker();   
  });

  it('handles valid request', function(done) {
    request(app)
      .post('/message')
      .send({version: 10, docId: 'blah'})
     .expect(200)
     .end(done);
  });
});
```

This test passes the first time every time. Then after about 3 file saves it will fail for the next few file saves, and eventually pass again intermittently going back and forth with the double callback error.

```
TypeError: Cannot read property 'length' of undefined
    at Test.Request.callback ([project]/node_modules/supertest/node_modules/superagent/lib/node/index.js:746:14)
    at ClientRequest.<anonymous> ([project]/node_modules/supertest/node_modules/superagent/lib/node/index.js:711:10)
    at ClientRequest.EventEmitter.emit (events.js:95:17)
    at Socket.socketOnEnd [as onend] (http.js:1568:9)
    at Socket.g (events.js:180:16)
    at Socket.EventEmitter.emit (events.js:117:20)
    at _stream_readable.js:919:16
    at process._tickDomainCallback (node.js:463:13)
```

lukebayes:
I may have found the root cause and a workaround.

It looks like providing a callback handler to `expect` _and_ adding an `end` handler may intermittently trigger this multiple callback warning.

I have changed our test code to always pass a callback to `expect`, never call `end()` and the provided handler calls `done()` directly.

Haven't seen a recurrence for an hour or so.

JohnTigue:
I found my way here because of "double callback!" messages while running Mocha tests with Nock. I am leaving this message mostly to cross-link to more info at #417.

And in Nock's issues: https://github.com/pgte/nock/issues/211

One small trick for checking to see if the SuperAgent-to-Nock interaction is your problem is to call nock.enableNetConnect() right before firing off a request with SuperAgent. Toggle that on and off with comments while running Mocha and you might just have found where your prey is hiding.

SGrondin:
I have found a workaround for my particular case.

In a mocha test file, I was doing:
```javascript
[...]

PR #1005, opened 2016-06-17
fix for nock compatibility regarding callback being invoked twice

samouri:
This is a fix for one of the common cases of the double-callback issue
#417
#313

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -536,6 +536,8 @@ Request.prototype.request = function(){
     // if we've recieved a response then we don't want to let
     // an error in the request blow up the response
     if (self.response) return;
+    // if end has already invoked the callback
+    if(self.called) return;
     self.callback(err);
   });
 


focusaurus:
How about reproducing the issue in a unit test and thus verifying the fix?

pornel:
We already have a check that prevents callback from being called twice:

https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L575

which is exactly what prints "double callback", because the message is not a bug, but a diagnostic that warns about presence of other logic bugs.

So bypassing "double callback" check doesn't fix multiple callbacks, it just hides them.

samouri:
@pornel, what is the desired behavior in this situation:
1. client of request library creates a new request with various params
2. client calls .end()
2.1 within .end(), there is an error with adding a the query params.  This causes the callback to be called: https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L642-L659
2.2 The http request created in end() throws an error.  As things are now this will cause the callback to be called again: https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L531-L540

pornel:
I'm afraid nock changes behavior here significantly. 

In superagent this sitation can't happen, because the request is made _after_ `_appendQueryString`, so it can't fire an error, because it's not started yet until end of the function: https://github.com/visionmedia/superagent/blob/86921996137a40b6fa12d343ca0d32daa2993fc6/lib/node/index.js#L844

samouri:
So let me make sure I understand this correctly: 
Superagent would never run into this normally because _Node.js_'s [Http.ClientRequest](https://nodejs.org/api/http.html#http_class_http_clientrequest) does _not_ emit errors until the request has been started (`flushHeaders`, `write`, `end`, etc. has been called).

For this I agree that nock should not emit errors until one of the request-starting functions have been called. I'll open up a thread with them on the topic.

But in terms of superagent handling requests -- wouldn't it make sense to protect against this sort of issue (clients directly accessing the request object or tampering with it somehow like in the case of nock) by clearing listeners to `req` any time `superagent.end` aborts mid-way (like in this case of [errors adding query string](https://github.com/visionmedia/superagent/blob/86921996137a40b6fa12d343ca0d32daa2993fc6/lib/node/index.js#L656))

pornel:
You're correct.

And yes, it sounds like a good idea to remove listeners (or maybe not even add them at that point yet) if something goes wrong in the setup.

pornel:
I think moving appendQueryString (but not its catch) to inside `.request()` before handlers are attached would solve the problem, and remove some duplication.

samouri:
I looked into it and it would break superagent functionality.  Usually people can call `.query` multiple times before sending out a request and each new query gets appended.    The way `appendQueryString` is right now, it is meant to be a final mutator.  It will only append if there is no querystring yet.

pornel:
We did some refactoring since, and this change is still not what we want, so I'm closing this.

PR #1007, opened 2016-06-21
enable field() to handle an object

affanshahid:
Allows `.field()` to behave similar to `.set()` and handle an object to set multiple key-values. Would it also be desirable to have it handle arrays? i.e
The following:
```
req.field({ array: [1, 2, 3]  });
```
would translate to something like:
```
req.field('array', 1);
req.field('array', 2);
req.field('array', 3);
```

The underlying `form-data`'s `append()` does not support this and errors out as it feels this sort of call is ambiguous (multiple interpretations). Would this be inline with superagent or is it indeed too ambiguous? 

Also I could not find browser-side tests for `.field()`, is this intended?

### DIFF
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -166,21 +166,31 @@ exports.unset = function(field){
 };
 
 /**
- * Write the field `name` and `val` for "multipart/form-data"
- * request bodies.
+ * Write the field `name` and `val`, or multiple fields with one object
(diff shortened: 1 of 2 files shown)


pornel:

Consistency with set is a good idea.

We're having problems with `localtunnel.me`, so I'm going to give Travis a few kicks and merge it when tests pass.

PR #1023, opened 2016-07-08
request-base: add header for inspect

magicdawn:
header is also important for inspect a request. as I'm using the `Authentication` header for auth. and I'm using this

```js
request
  .get(url)
  .set('foo', 'bar')
  .then((err, res) => {
    if(err) return next(err);
  });
```

`next` is the `next` in express. I'd like to inspect the header too.

### DIFF
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -247,7 +247,7 @@ exports.toJSON = function(){
     method: this.method,
     url: this.url,
     data: this._data,
-    header: this._header
+    headers: this._header
   };
 };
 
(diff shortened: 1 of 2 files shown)


pornel:
Should it be `header` or `headers`?

magicdawn:
@pornel That's a question ...

pornel:
In node it's `headers`, so I suggest going with that: https://nodejs.org/api/http.html#http_message_headers

Also, could you add a basic unit test for it?

magicdawn:
Oops ...

pornel:
Thank you

PR #1027, opened 2016-07-13
Update jQuery

pornel:


focusaurus:
Ha you're too fast! I was trying to replace jquery-ui with the latest in the repo but their custom package builder server is throwing 504s at the moment. This looks good to me and getting the libs from CDN should be nice.

PR #1029, opened 2016-07-18
Skip null querystring values in node environment.

Gerhut:
In order to match the behavior of browser environment.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -70,9 +70,7 @@ function serialize(obj) {
   if (!isObject(obj)) return obj;
   var pairs = [];
   for (var key in obj) {
-    if (null != obj[key]) {
-      pushEncodedKeyValuePair(pairs, key, obj[key]);
-    }
+    pushEncodedKeyValuePair(pairs, key, obj[key]);
   }
   return pairs.join('&');
 }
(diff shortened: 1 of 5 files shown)


pornel:
Thanks. Consistency is definitely desirable. However, maybe we should adjust browser to match node instead?

`JSON.stringify` keeps `null`s:

```js
JSON.stringify({n:null,u:undefined})
'{"n":null}'
```

Gerhut:
I prefer ignoring them because `{"a": null}` compiles `a=`, which may not be decided with `{"a": ""}`.

pornel:
Query string taking JSON isn't meant to preserve it 100% (for that you'd use actual JSON post type), it's just a convenience. I think it's fine as long as it's consistent and seems sensible.

Perhaps we could serialize `{a:null,b:1}` as `?a&b=1`.

pornel:
@focusaurus what do you think about it?

focusaurus:
I like your suggestion of `{a:null,b:1}` encodes to `?a&b=1`. That seems reasonable.

More broadly, I dislike informal serialization schemes that try to map nested objects and arrays into a query string since they tend to be poorly-defined and have broken edge cases. If I were to rethink the superagent API I might explicitly limit `.query(foo)` to a single non-nested object of names and scalar values.

pornel:
Thank you @Gerhut!

PR #1037, opened 2016-08-05
Be strict with multiple types of body in single request

rafis:
Based on issue:
https://github.com/visionmedia/supertest/issues/356


pornel:
I'm wondering, instead of making it fail, can we make it just work instead?

I guess it'd require using `this._data` for `.field` and `.attach`, and then "serializing" it as `FormData` before sending.

PR #1038, opened 2016-08-07
The returned object is not compatible with Promises.

Nitive:
We can't only catch errors for request
```js
const superagent = require('superagent')

superagent
  .get('https://ya.ru')
  .catch(err => console.log(err))
```

Output:
```
  .catch(err => console.log(err))
        ^

TypeError: superagent.get(...).catch is not a function
```

But in promises we can:

```js
Promise.reject('Error!')
  .catch(err => {
    console.log(err) // Error!
  })

Promise.resolve(123)
  .catch(err => {
    // no errors
    console.log(err)
  })
```

---

Added test to reproduce error. I can try to fix it if you are ok with it.

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -384,6 +384,22 @@ describe('request', function(){
     })
   })
 
+  describe('.catch(reject)', function() {
+    it('should reject an error with .catch(reject)', function(done) {
+      if ('undefined' === typeof Promise) {
+        return done();
+      }
+
+      request
+      .get(uri + '/error')
+      .catch(function(err) {
+        assert(err.status == 500);
+        assert(err.response.text == 'boom');
+        done();
+      })
+    })
+  })
+
   describe('.abort()', function(){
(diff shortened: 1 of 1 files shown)


focusaurus:
So the superagent API returns the request instance you are building up, not a promise, from most of the API methods. This is because the superagent API is based on the chaining paradigm. In order to get a promise, you must call `.then()` which deviates from the main pattern in that it returns a promise, not the request instance. Once you have the promise, you can call its `.catch()` method.

We need the API to work this way so we can tell when you are done configuring the request and thus we should actually send it vs you plan to further chain calls to set headers, etc.

pornel:
Technically superagent is a "thenable" and doesn't require to have `catch`, but I see your point, and since the implementation only takes 1 line, I think it's fine to add it.

Thanks for the PR.

Nitive:
Thanks!

PR #1041, opened 2016-08-11
Set xhr.response instead of xhr.responseText to err.rawResponse

kolodoz:
responseText throws InvalidStateError when responseType is not 'text' or ''


pornel:
IIRC `.response` is an XHR2 feature, which IE9 does not support.

kolodoz:
Sorry, forgot about our old friend. Fixed and added a test.

kolodoz:
Missed the button

focusaurus:
LGTM. @pornel should we merge this?

PR #1042, opened 2016-08-11
Fix calling abort before XHR setup

stephanebachelier:
Fix runtime issue in 1.8.x branch when calling `abort` before the XHR have been created.


pornel:
Thanks. What's your reason for staying on 1.8?

stephanebachelier:
@pornel Thanks! 

I've some legacy code and I don't have the time to upgrade to 2.x.x (for now!)

PR #1047, opened 2016-08-15
Return this when overwriting the response object's on() method

juandopazo:
Superagent is overriding Node's `res.on` function, but forgetting to return `this` to maintain the original behavior.


pornel:
Thanks

juandopazo:
Awesome! Thank you!

PR #1048, opened 2016-08-16
guard for in when setting headers

stonerworx:
We're having an issue with IE11 that is trying to set a `Symbol(Symbol.iterator)` as a header. Guarding the `for .. in` loop fixes this issue. Do you think it makes sense implementing this?
Thanks!

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -825,8 +825,9 @@ Request.prototype.end = function(fn){
 
   // set header fields
   for (var field in this.header) {
-    if (null == this.header[field]) continue;
-    xhr.setRequestHeader(field, this.header[field]);
+    if ({}.hasOwnProperty.call(this.header, field)) {
+        xhr.setRequestHeader(field, this.header[field]);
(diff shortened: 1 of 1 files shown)


pornel:
it makes sense. What about the null check?

stonerworx:
I closed the pull request as it was actually not the reason for our problem. Still it would make sense to implement. Yes, the null check should be kept.

lib/client.js
```diff
@@ -825,8 +825,9 @@ Request.prototype.end = function(fn){
 
   // set header fields
   for (var field in this.header) {
-    if (null == this.header[field]) continue;
-    xhr.setRequestHeader(field, this.header[field]);
+    if ({}.hasOwnProperty.call(this.header, field)) {
```
focusaurus: Would it be marginally more efficient to do `Object.prototype.hasOwnProperty` here instead of declaring a new object literal each turn of the for loop?

Issue #977, opened 2016-04-27
how to compute progress percent of request in superagent


shreychaturvedi123:

pornel:
The `progress` event gives an object as an argument which includes information like amount downloaded and total size.

shreychaturvedi123:
can you show me a snippet?

pornel:
```js
superagent.get('https://example')
.on('progress', function(e){
   console.log(e.direction,"is done",e.percent,"%");
})
.end();
```

Note that we support *download* progress only in browser, and *upload* progress only in node (it's weird, I know).

fisch0920:
@pornel where in the world is this documented?? I've been looking through docs for the past few hours and just happened to stumble on this issue which is the only place I've clearly found this distinction.

Why isn't upload progress supported in the browser?

pornel:
Documentation for it is missing. I'd be helpful if you made a pull request adding it!

Upload progress in browsers is just not implemented. I think with XHR2 it should be possible to add it.

fisch0920:
Okay; I'd be glad to look into adding support, but I'm not sure I understand why it doesn't work as-is, given that `lib/client.js` includes a progress handler and bindings for `xhr.onprogress` and `xhr.upload.onprogress`.

Do you have any insight or resources that would explain why this approach isn't currently working?

Thanks!

fisch0920:
Also, it looks like the progress handler is called several times on Chrome in a few tests I've run, but only at like 98% and 100% and never consistently.

halt-hammerzeit:
@fisch0920 So, does browser file upload progress work?
It's surely supported, and has been so for a long time.

pornel:
Please note that for small files you may not get any events or get nonsense progress. That's simply due to how TCP/IP and buffering works — until all buffers in your browser/os/router are full, the sending appears to be instant, but it's not known how much reached the other end until there's a response from the other side.

PR #1049, opened 2016-08-17
Correct event.direction in uploads

pornel:
I didn't realise we had upload progress, it was just buggy.

#977


fisch0920:
Awesome -- thanks @pornel!

PR #1056, opened 2016-08-29
Indicate which versions superagent-cache can use

jpodwys:
Superagent-cache is not yet compatible with superagent `2.x`. Adding a note so people using `2.x` aren't mislead.


focusaurus:
Thanks!

PR #1060, opened 2016-09-02
Remove superagent-promise-plugin from plugin list

pushred:
Support added in 2.0.0 negates any utility and it's presence in this list implies that Promises are _not_ supported. This was pretty confusing to me, even though the docs on the site clearly communicate that Promises are now supported. I look at READMEs first and docs elsewhere a distant second.

I also somehow missed that news earlier this year and caught myself using thenables after forgetting to install the plugin in a project. Retracing my steps wondering why it was actually working, I discovered the good news!


pornel:
Good catch.

Issue #1065, opened 2016-09-09
req.field('key', 'value') is broken


magicdawn:
the 2.0.0 release use form-data@1.0.0-rc4
while will cause 

```js
TypeError: Cannot read property 'path' of null
    at FormData._getContentDisposition (/.../node_modules/superagent/node_modules/form-data/lib/form_data.js:195:43)
    at FormData._multiPartHeader (/.../node_modules/superagent/node_modules/form-data/lib/form_data.js:169:33)
    at FormData.append (/.../node_modules/superagent/node_modules/form-data/lib/form_data.js:64:21)
    at Request.exports.field (/.../node_modules/superagent/lib/request-base.js:184:23)
```

the call chain
- https://github.com/visionmedia/superagent/blob/v2.2.0/lib/request-base.js#L184
- https://github.com/form-data/form-data/blob/v1.0.0-rc4/lib/form_data.js#L64
- https://github.com/form-data/form-data/blob/v1.0.0-rc4/lib/form_data.js#L169
- https://github.com/form-data/form-data/blob/v1.0.0-rc4/lib/form_data.js#L195

magicdawn:
`value.length` <- the exception

pornel:
That looks like a bug in `FormData` (I don't see what we can do about it in superagent). Can you report the problem there?

magicdawn:
I'm sorry, Just found that error happens when the `value` is empty. 
And can be reproduced by code below: 

`.field('key')`, `value` is `undefined`

```js
'use strict';

const request = require('superagent');
const co = require('co');

const main = co.wrap(function*() {
  const res = yield request
    .post('http://localhost')
    .field('x');
});

if (module === process.mainModule) {
  main().catch(e => console.error(e.stack || e));
}
```

So should we add a `assert(value, 'value can not be empty')` here or the form-data repo ?

magicdawn:
what does an empty field means in a `multipart/form-data` ?
Looks there already got some clues

- form-data/form-data#147
- request/request#2208

pornel:
`.set('header',undefined)` throws, `.query('field',undefined)` sets query string to `?field`, so it could go either way.

Throwing may be easier, but making undefined remove fields from the form could be more useful.

magicdawn:
Since form-data is going to fix this in v2, and current this will throw a unknown error
`TypeError: Cannot read property 'path' of null`, and I found a hidden bug in my bussiness code after figuring out what really happens. 😂

[...]

PR #1066, opened 2016-09-12
req.field assert

magicdawn:
#1065 related


magicdawn:
I noticed that the test code style are not consistent in semi `;` http://eslint.org/docs/rules/semi

pornel:
We're not using assert module in the client code. Can you use regular if/throw instead?

pornel:
Superb. Thanks!

PR #1071, opened 2016-09-18
Error event

pornel:


lib/client.js
```diff
@@ -689,6 +689,9 @@ Request.prototype._getFormData = function(){
 Request.prototype.callback = function(err, res){
   var fn = this._callback;
   this.clearTimeout();
+
+  this.emit('error', err);
```
focusaurus: At a glance this looks like it will emit the error event even on successes when error is null. Am I missing something?

focusaurus:
Test failures look like legitimate issues that should be addressed, not saucelabs flakiness.

pornel:
Yup

PR #1072, opened 2016-09-19
Add the response to an error inside a promise

DavidTanner:
Sometimes the res object contains critical data as to why a request failed.  When a promise errors out, it would be nice to have that information as part of the error object.


pornel:
There is a `.response` field already

DavidTanner:
https://github.com/visionmedia/superagent/blob/v2.2.0/lib/client.js#L503
Cool, I didn't see that.  Thanks

Issue #1079, opened 2016-10-06
`Request#attach()` should allow access to more options


lapo-luchini:
#520 added an optional `filename` parameter, allowing access to only one of the underlying options, but it would be useful to use other at times, as sometimes a specific mime-type is required (and pre-known length might help too).

I suggest accepting an object in that parameter and use it straight as an `options` object, while keeping the current `{filename: param}` when a string is given.

E.g. [FormData example](https://github.com/form-data/form-data):

```javascript
form.append('file', stdout, {
    filename: 'unicycle.jpg',
    contentType: 'image/jpg',
    knownLength: 19806
});
```

pornel:
That's a good idea. It may need some care to keep compatibility with the browser. 

Would you make a PR for this?

lapo-luchini:
I'm very willing to PR for the node version (as that's what I'm currently using), but I'm not sure when I'd be able to check it in the browser too.

pornel:
AFAIK in browser there's no control beyond filename, so it'd be fine to ignore extra parameters. As long as you don't break 

```sh
make test-browser-local
```

it's fine :)

lapo-luchini:
Oh, I noticed that [in the browser code](https://github.com/visionmedia/superagent/blob/master/lib/client.js#L651) while the third field is called `filename` the example it's already using it as an object to specify the content type (though it seems strange that it's using `type` instead of `contentType` [as accepted by the library](https://github.com/form-data/form-data/blob/master/lib/form_data.js#L234); maybe there is a conversion somewhere in the code path, but I didn't find it).

Also, the browser tests won't break, as [the only test](https://github.com/visionmedia/superagent/blob/master/test/client/request.js#L64) doesn't cover the third parameter.

PR #1081, opened 2016-10-10
Support the full `attach` options parameter in the node version too.

lapo-luchini:
As requested in #1079.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -650,7 +650,7 @@ Request.prototype.query = function(val){
 
 /**
  * Queue the given `file` as an attachment to the specified `field`,
- * with optional `filename`.
+ * with optional `options` (or filename).
  *
  * ``` js
  * request.post('/upload')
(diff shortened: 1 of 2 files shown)


lapo-luchini:
(I'm not sure about travis, but it didn't break `make test-browser-local`)

martinbooth:
Hey. What is `filename` in https://github.com/visionmedia/superagent/blob/b179a6b3766b8f322b1d3bd91469c4755e190dc9/lib/node/index.js#L185 ? Doesn't seem to be defined anywhere is it?

lib/client.js
```diff
@@ -660,13 +660,13 @@ Request.prototype.query = function(val){
...
- * @param {String} filename
+ * @param {String|Object} options
  * @return {Request} for chaining
  * @api public
  */
 
-Request.prototype.attach = function(field, file, filename){
-  this._getFormData().append(field, file, filename || file.name);
+Request.prototype.attach = function(field, file, options){
+  this._getFormData().append(field, file, options || file.name);
```
lapo-luchini: No, as the [third parameter is an object](https://github.com/form-data/form-data/blob/19748778e2cd05e2ddc95409fcbbf0461a868775/lib/form_data.js#L44), even though a string is accepted.

lapo-luchini:
@martinbooth you're right, that should be a `o.filename`.

PR #1082, opened 2016-10-11
Remove componentjs mentions, update Readme format

damiencaselli:
Component.js is no longer maintained (see componentjs/component#639).


pornel:
Thanks

PR #1087, opened 2016-10-19
Add component-emitter to bower.json

NullSoldier:
The library relies on component-emitter but is not listed in the bower.json.

There are a few others too, just haven't run into issues yet.


pornel:
Thanks.

PR #1088, opened 2016-10-21
Mixin RequestBase functions from a prototype

ahelmberger:
This PR is an alternative solution to #1076

* Make RequestBase a constructor function
* Make it mixin its prototype functions when called as a function
* Unifies mixin approach with Emitter
* Fixes global scope issues in module bundlers like rollup.js

References #1076

See also:
https://github.com/rollup/rollup/issues/1007
https://github.com/rollup/rollup/pull/1023
https://github.com/rollup/rollup-plugin-commonjs/issues/127

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -13,7 +13,7 @@ if (typeof window !== 'undefined') { // Browser window
 }
 
 var Emitter = require('emitter');
-var requestBase = require('./request-base');
+var RequestBase = require('./request-base');
 var isObject = require('./is-object');
 
 /**
@@ -517,13 +517,11 @@ function Request(method, url) {
 }
 
 /**
- * Mixin `Emitter` and `requestBase`.
+ * Mixin `Emitter` and `RequestBase`.
  */
 
 Emitter(Request.prototype);
-for (var key in requestBase) {
-  Request.prototype[key] = requestBase[key];
-}
+RequestBase(Request.prototype);
 
 /**
  * Set Content-Type to `type`, mapping values from `request.types`.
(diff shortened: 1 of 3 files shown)


ahelmberger:
Red build on TravisCI doesn't seem to be a code issue.

pornel:
Thank you for the PR. That looks like a good way to solve it.

I wonder if we need to bump semver-major for this? (will this impact supertest, plugins or mocking?) WDYT @focusaurus?

focusaurus:
It might be worth getting some explicit coordination from the supertest maintainers on this branch before merging. I believe their mechanism to integrate with superagent will continue to work but I think it would be worthwhile to test. Looks like @mikelax is the most recent active maintainer. Perhaps I could try to pull down the supertest repo, integrate it with this branch of superagent, and see if their unit tests still pass.

focusaurus:
I was able to:

- clone supertest
- npm install, run tests, they pass
- `rm -rf ./node_modules/superagent && npm i 'ahelmberger/superagent.git#mixin-requestbase-prototype'`
- supertest unit tests still pass

So that's a good sign.

focusaurus:
Pragmatically I don't think there's much harm done in treating this as semver major even if in retrospect we find out nothing important broke. That's probably better than semver minor that unexpectedly breaks some integration, so I wouldn't sweat it too much.

pornel:
Cool. Let's do that as semver-major. It's also a good time to drop Node 0.10 support (official support has ended), which might be seen as semver-major too.

mindrones:
Hi, this patch is [in master](https://github.com/visionmedia/superagent/blob/master/lib/request-base.js) but not in the [latest release](https://github.com/visionmedia/superagent/blob/v2.3.0/lib/request-base.js): would it be possible to release again so that superagent can be properly used in projects built with Rollup? Thanks!

PR #1089, opened 2016-10-22
Added random User Agent method to utils

pxai:
I added this simple function for me. Maybe somebody finds it useful.
Applies a random User Agent instead of a fixed one or the default _node-superagent_.

Example of use:

```
request.get(url)
        .set('User-Agent', utils.randomUa())
        .end(function(err, res){
            if (err) throw err;
            console.log(res.text);
        });
```

### DIFF
--- a/lib/node/utils.js
+++ b/lib/node/utils.js
@@ -134,7 +134,6 @@ exports.unzip = function(req, res){
  * @return {Object} header
  * @api private
  */
-
 exports.cleanHeader = function(header, shouldStripCookie){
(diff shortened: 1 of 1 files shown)


focusaurus:
Nice! I think this might be more appropriate as a superagent plugin instead of in the core module.

pornel:
Thanks for the pull request, but I think it's a bit too specific functionality to be included in a general library like superagent. That may be useful as a standalone module or maybe a function in supertest.

PR #1097, opened 2016-11-03
Initialize request data

frmendes:
I recently had an issue where I would be making two consecutive POST request. Both received an object as `data` and both used a second `csrf` object. Something like:

```javascript
var csrf = extractCSRF();

request1 = (data) => {
  return new Promise((resolve, reject) => {
    request
      .post(ENDPOINT_1)
      .set(HEADERS)
      .send(csrf)
      .send(data)
      .end((err, res) => err ? reject(err) : resolve(res.body));
  });
}

request2 = (data) => {
  return new Promise((resolve, reject) => {
    request
      .post(ENDPOINT_2)
      .set(HEADERS)
      .send(csrf)
      .send(data)
      .end((err, res) => err ? reject(err) : resolve(res.body));
  });
}
```

The second request would fail because of conflicting parameters. The `csrf` object was being mutated due after the first request causing a merge conflict and a malformed request.

I was able to track this down to [this line](https://github.com/visionmedia/superagent/blob/master/lib/request-base.js#L379). This check would fail because `this._data` was uninitialized, meaning that the object cloning wouldn't occur. Instead, it was falling down to the `else` branch which contains the line `this._data = data`. As a consequence, `send` was mutating the `csrf` object which is unwanted behavior.

I fixed this issue by initializing `this._data`.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -474,6 +474,7 @@ function Request(method, url) {
   this._query = this._query || [];
   this.method = method;
   this.url = url;
+  this._data = {};
   this.header = {}; // preserves header name case
   this._header = {}; // coerces header names to lowercase
   this.on('end', function(){
--- a/lib/client.js
+++ b/lib/client.js
@@ -474,7 +474,6 @@ function Request(method, url) {
   this._query = this._query || [];
   this.method = method;
   this.url = url;
-  this._data = {};
   this.header = {}; // preserves header name case
   this._header = {}; // coerces header names to lowercase
   this.on('end', function(){
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -374,10 +374,11 @@ RequestBase.prototype._isHost = function _isHost(obj) {
 RequestBase.prototype.send = function(data){
   var obj = isObject(data);
   var type = this._header['content-type'];
+  var objectType = {}.toString.call(data);
 
   if (obj && Array.isArray(data)) {
     this._data = this._data || [];
-  } else if (obj && 'object' == typeof data) {
+  } else if (obj && objectType == '[object Object]') {
     this._data = this._data || {};
   }
 
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -375,6 +375,12 @@ RequestBase.prototype.send = function(data){
   var obj = isObject(data);
   var type = this._header['content-type'];
 
+  if (obj && Array.isArray(data)) {
+    this._data = this._data || [];
+  } else if (obj && 'object' == typeof data) {
+    this._data = this._data || {};
+  }
+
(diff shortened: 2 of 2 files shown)


pornel:
I agree there's a bug. Superagent should be copying given objects instead of mutating them.

Unfortunately this particular workaround is breaking things:

> Uncaught AssertionError: expected '{"0":1,"1":2,"2":3}' to be '[1,2,3]' 

Please check

```sh
make test-browser-local
```

frmendes:
Well, my fix was based on the wrong assumption that `_data` would be an object every time. Instead it should only be initialized to `{}` when the `data` variable on [`RequestBase.prototype.send`](https://github.com/visionmedia/superagent/blob/master/lib/request-base.js#L374) is an actual object. It should also account for Arrays, which I believe lead to a similar situation.

I have the following workaround:

```javascript
RequestBase.prototype.send = function(data){
  var obj = isObject(data);
  var type = this._header['content-type'];

  if(obj && data.constructor === Object) {
    this._data = this._data || {}
  } else if (obj && data.constructor === Array) {
    this._data = this._data || []
  }

  // ...
```

`obj` is true for Arrays which leads to the error you mentioned. So I had to add `data.constructor` part. I'm on the fence with putting this logic here. What do you think?

pornel:
1. There's `Array.isArray`
2. Why `|| {}` and `|| []` if you know the object won't be falsy?

frmendes:
1. I wanted to keep consistency because of `data.constructor === Object` but I can switch it to `Array.isArray`
2. I'm not sure I'm getting your question but `obj` refers to `data` not being falsy, not `this._data`. There are times where `this._data` is initialized and it avoids overriding it.

pornel:
1. Constructor check is weird. Correctness is more important, so if you check for `isArray` first, and then for `typeof`, it'll work the best
2. Sorry, I've misread the code!

frmendes:
I agree with you with the constructor check. I pushed a version following your suggestion but it's failing for [buffer object data](https://github.com/visionmedia/superagent/blob/master/test/node/basic.js#L261).

I can check for the content-type, since it's only really an issue with JSON, right? Or use the uglier constructor check which doesn't seem to cause these issues.

pornel:
Oops, sorry. My suggestion with typeof object was wrong. The input can also be a `Buffer`/`Blob`, etc.

Test failure: https://travis-ci.org/visionmedia/superagent/jobs/173242438

You can use `npm test` to check these locally.

frmendes:
I updated to explicitly look for an object. Not sure if it's the correct way, but seems better than to use a constructor check.

pornel:
OK, now it all works. Thank you.

Issue #12, opened 2011-08-22
make: Nothing to be done for `all'.


sarendipitee:
Running make doesn't do anything. How do I build this darn'd thing? O.o

[JD-i7-iMac:superagent JD]$ make -f Makefile 
make: Nothing to be done for `all'.

Just so we're on the same page... this IS a script for making ajax calls in the browser, right? i.e., not just in node?

PR #1102, opened 2016-11-07
Support `pfx` TLS option with new pfx() method.

reid:
This change introduces the pfx() method for setting the `pfx` TLS option. This change was last attempted in #685 but that PR did not have tests.

https.request() supports the TLS options of tls.connect(). The [`pfx` option][pfx] is a string or Buffer containing the private key, certificate, and CA certs of the client in [PKCS #12 format][p12].

Agent is modified to also accept the `pfx` option.

The `cert.pfx` file is used in testing instead of the CA and certificate files to make sure requests work correctly when presenting the PFX archive instead of its parts. Since the server is uses the same CA and certificate fixtures, the content of `cert.pfx` is derived from these files using the command:

    openssl pkcs12 -export -out cert.pfx -inkey key.pem -in cert.pem -certfile ca.cert.pem

[pfx]: https://nodejs.org/dist/latest-v6.x/docs/api/tls.html#tls_tls_createsecurecontext_options
[p12]: https://en.wikipedia.org/wiki/PKCS_12


pornel:
Thank you for the PR, with the tests 👍

a-ludi:
It would be nice to have some way of providing the `passphrase` for the cert/key file. ATM I can't prepare a PR myself, so I am writing at this place, hoping one of you guys can do that.

PR #1103, opened 2016-11-08
saucelabs browser test tweaks

focusaurus:
- 20s timeouts
- latest zuul
- configure SAUCE_APPIUM 1.6

I was seeing our 10s timeouts exceeded a lot in sauce labs for slower browsers/OSes. Increasing to 20s seems empirically to help. :-/

### DIFF
--- a/.zuul.yml
+++ b/.zuul.yml
@@ -1,5 +1,6 @@
 ui: mocha-bdd
 server: ./test/support/server.js
+tunnel_host: http://focusaurus.com
 browsers:
   - name: chrome
     version: latest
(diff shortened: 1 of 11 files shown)


focusaurus:
Well, most of the sauce browser tests actually passed. The one that failed has:

```
<internet explorer 10 on Windows 2012> request :: on 303 :: should redirect with same method
Error: Unable to get property 'text' of undefined or null reference (http://jhpwkkjpxw.localtunnel.me/__zuul/test-bundle.js:10232)
```

Which I would think means localtunnel failed to serve the test-bundle properly.

focusaurus:
Maybe let's try focusaurus.com (a domain I bought just to run localtunnel-server in a docker container after realizing localtunnel-server only works on top-level hostnames), at least for a while to see if there really are reliability problems with localtunnel.me

pornel:
That's great. Thank you.

pornel:
And it's cool that there are still .com domains available :)

PR #1108, opened 2016-11-15
README - query string

oprogramador:


focusaurus:
It's not clear to me why the plugin example needs updating this way. Perhaps you did not encounter `docs/index.md` which powers the main superagent website and documentation?

oprogramador:
I think it's an essential example and it should be in README.md (at least `query`).

pornel:
Query example seems like a good thing to add.

While we're at it, could we have an example with some realistic values, not "foo" and "some-url"?

oprogramador:
@pornel I updated this PR.

pornel:
Great! thank you

Issue #859, opened 2016-01-18
Using both `attach` and `send` incorrectly results in multipart data being sent as JSON


joepie91:
While helping out @devJunkORG on IRC with a strange issue, we ran into this issue: when using both `.attach` and `.send` in a request, the request header will indicate `application/json` as `Content-Type`, but the payload will contain multipart-formatted data (with boundaries and all).

The following is example code:

``` js
files.forEach((file) => {
    req.attach(file.name,file);
});

req.send({ id: 12345 })

req.end((err,res) => {
    // ...
});
```

The following is a (redacted) snippet of the HTTP request that is made:

```
POST /upload HTTP/1.1
Connection: upgrade
Host: <redacted>
Content-Length: 476885
Origin: <redacted>
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.111 Safari/537.36
Content-Type: application/json
Accept: */*
Referer: <redacted>
Accept-Encoding: gzip, deflate
Accept-Language: es-ES,es;q=0.8,en-US;q=0.6
Cookie: _<redacted>

------WebKitFormBoundaryahUn9Gpiz9X7ii4e
Content-Disposition: form-data; name="sample.png"; filename="undefined"
Content-Type: image/png
```

The cause of this issue is that superagent does not prevent a user from using both `attach` and `send`, even though that is invalid and the user should be using `field` instead. Thus, the following happens:
1. [`this._formdata` is set](https://github.com/visionmedia/superagent/blob/8b10519376114ee5df6e0c77ef6adc77992f1259/lib/client.js#L793-L794)
2. [`this._data` is set](https://github.com/visionmedia/superagent/blob/8b10519376114ee5df6e0c77ef6adc77992f1259/lib/client.js#L858)
3. The `content-type` [is set to `application/json`](https://github.com/visionmedia/superagent/blob/8b10519376114ee5df6e0c77ef6adc77992f1259/lib/client.js#L862), leading to the header indicating JSON.
4. When sending, [`this._formData` is prioritized over `this._data`](https://github.com/visionmedia/superagent/blob/8b10519376114ee5df6e0c77ef6adc77992f1259/lib/client.js#L941), leading to the payload containing multipart-encoded data.

[...]

PR #1113, opened 2016-11-19
Don't allow unsupported mixing of data types

### LINKED ISSUE #209: Mixed form data and attachments don't construct valid message
I'm having trouble mixing form data with file submissions - a valid form of multipart form data that is used in certain apis doesn't seem to be testable.  I put together a super simple test that demonstrates the issue here:

https://github.com/DavidIAm/SuperTestFailExample

It is my best guess the flaw that allows this behavior is in superagent, though its possible its in supertest and also possible its in the general http library code.  By all means inform and I will escalate the issue appropriately.

pornel:
Closes #209

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -671,6 +671,10 @@ Request.prototype.query = function(val){
  */
 
 Request.prototype.attach = function(field, file, options){
+  if (this._data) {
+    throw Error("superagent can't mix .send() and .attach()");
+  }
+
   this._getFormData().append(field, file, options || file.name);
   return this;
 };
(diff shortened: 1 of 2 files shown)


DavidIAm:
...are you suggesting that the use case of sending a multipart response with some form feilds and some file attachments is one for which there is so little utility that the superagent should not allow it? I am confused because not only is this is a legal and supported by traditional user agents form of multipart message ( part as www form data, another part as a file content ) but some file put apis absolutely require this in order to work, sending metadata in the form fields along with the file. We will not be able to replicate the beahavior of a normal web browser file submit if we cannot include form fields in the upload!

pornel:
No, it's that `.send()` never ever worked together with `.attach()`. It was either silently dropping data or sending garbage, so the only change is that now you'll know about it.

`.attach()` can only work with `.field()`.

devjunkORG:
Also closes #859

PR #1120, opened 2016-12-02
DRY Up Response in Node and Browser

leipert:
Recently we were wondering why parsed `response.links` can be found in the node version of superagent, while not being there in the browser version, after a quick debugging session we found the reason why. In this PR we did the following things to get more parity between node and browser versions:

1.  split `lib/node/utils` into `lib/utils` and `lib/node/unzip`, so that `lib/utils` can be used in the browser code and does not bring unwanted dependencies (namely `unzip`) into the browser.

    Maybe `lib/node/unzip` should be renamed to `lib/node/node-utils` or extend `lib/utils`, but this seemed to be the cleanest solution at the moment.

2.  After that we realized, that the client and node implementation of `Response` share a lot features and therefore moved the following shared functions of `Response` to a `response-base` similar to the `request-base`:

     - `get`
     - `_setHeaderProperties` (Finally we are getting our *precious* `links`)
     - `_setStatusProperties` (The IE9 status quirk was moved to the client Response constructor)

    Maybe the `Response.toError` function could be moved to `ResponseBase` as well, but that's something for another day :)
    
I hope that this PR satisfies your code requirements. If not please let me know and I could improve it further.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -16,7 +16,6 @@ var Emitter = require('emitter');
 var RequestBase = require('./request-base');
 var isObject = require('./is-object');
 var isFunction = require('./is-function');
-var utils = require('./utils');
 var ResponseBase = require('./response-base');
 
 /**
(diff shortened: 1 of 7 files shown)


pornel:
Thank you very much. This is good stuff!

PR #1126, opened 2016-12-07
Split .pipe() to avoid calling .end() second time after redirect

pornel:


lib/request-base.js
```diff
@@ -99,6 +99,9 @@ RequestBase.prototype.timeout = function timeout(ms){
 RequestBase.prototype.then = function then(resolve, reject) {
   if (!this._fullfilledPromise) {
     var self = this;
+    if (this._endCalled) {
+      console.warn("Warning: superagent request was sent twice, because both .end() and .then() were called. Never call .end() if you use promises");
+    }
```
focusaurus: I'd prefer curly braces for all if statements, but with or without is OK but let's make all 3 instances use the same syntax. The other 2 don't use curlies.

focusaurus: Otherwise LGTM.

Issue #1074, opened 2016-09-20
How to get message body for 4xx response? Or is it expected behaviour?


kroko:
Server is set up to return 200 and json object when server side validation and all further server work passes.
It returns 400 on server side validation error
It returns 500 on server side logic/DB/whatever error

Returning 400 is good praxis for server side validation failure (http://stackoverflow.com/a/34324179, http://stackoverflow.com/a/3290198, a.o.)
_Of course in real world there is also client side validation which forbids sending invalid form_

Example for 400

Client

``` javascript
import request from 'superagent';

...
    let sendToServerData;
    // sendToServerData = {one: 'isOne', two: 'isTwo'}; // test with data
    sendToServerData = {}; // test with no data

    request
    .post('/')
    .send(sendToServerData)
    .set('Accept', 'application/json')
    .then((sucess) => {
      console.log('we have sucsess', sucess);
    }, (faulure) => {
      console.log('we have failure', faulure);
    });
```

Server

``` php
if ($_SERVER['REQUEST_METHOD'] === 'POST') {

  // get input
  $rest_json = file_get_contents("php://input");
  // decode and put in post
  $_POST = json_decode($rest_json, true);

  // set header
  header_remove();
  header('Content-type:application/json;charset=utf-8');

  // if empty return 400 and json explaining things
  if(empty($_POST)) {
    http_response_code(400);
    header('Status: 400');
    exit(json_encode(array(
        'status' => 400,
        'message' => 'no empty posts my friend'
        )));
  }

  // just return back 200 and json explaining things
  http_response_code(200);
  header('Status: 200');
  exit(json_encode(array(
        'status' => 200,
        'message' => 'all fine!'
        )));

}
```

when `sendToServerData` is populated, all fine as expected - `sucess` contains all the payload. when `sendToServerData` is empty console prints `we have failure Error: Bad Request(…)` thus `faulure` object does not contain any server payload, it does not even contain status code. 

Is it expected (like "superagent way" would be returning 200 also on server side validation error, just set the payload to contain that there is actually server side validation error)? Or am I missing something? is it possibe to get object (`failure`) that contains server code and message on non 2xx status?

superagent@2.2.0

Thanks!

kroko:
I just tried dropping ES6 promises for this 

```javascript
    request
    .post('/')
    .send(sendToServerData)
    .set('Accept', 'application/json')
    .end((err, res) => {
      console.log(err, res);
    });
```
and it works - `res` body contains payload.

[...]

PR #1132, opened 2016-12-12
.ok()

### LINKED ISSUE #1069: Is there a way to config the 4xx & 5xx statusCode as normal response but not error
As the title presents. 

I'm tired of writing code like this

``` js
const getDataAsync = co.wrap(function*(){
  let res;
  try {
    res = yield request.get('some_url')
  } catch(err) {
    if(!err.original && err.response) {
      res = err.response
    } else {
      throw err;
    }
  }
});
```

I have to test `!err.original && err.response` exists, and treat it as success response, I'm tired of writing this. 
So I'm asking is there a way to config that and if not can this issue be a feature request?

### LINKED ISSUE #1096: 304 Not Modified http code request give this error
The first time that i make the request, it gives me an HTTP code `200`. But if i reload, it gives me a HTTP code `304 Not Modified` (which must be correct) and superagent throw an error.

`client.js:501 Uncaught (in promise) Error: Unsuccessful HTTP response(…)`

Is it correct!?

Version `1.7.1` works perfectly, but the latest gives me this error. (I rolled back, because i never saw this in the old version)

pornel:
Fixes #1069

Closes #1096 #1074

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -523,6 +523,14 @@ An "error" event is also emitted, with you can listen for:
       // all other error types we handle generically
     }
 
+Alternatively, you can use the `.ok(callback)` method to decide whether a response is an error or not. The callback to the `ok` function gets a response and returns `true` if the response should be interpreted as success.
+
+    request.get('/404')
+      .ok(res => res.status < 500)
+      .then(response => {
+        // reads 404 page as a successful response
+      })
+
 ## Promise and Generator support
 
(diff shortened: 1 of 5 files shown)


keithamus:
As someone interested in this issue, this gets a 👍 from me

focusaurus:
LGTM. Restarting travis to handle saucelabs browser test flakiness.

PR #1146, opened 2016-12-29
Update Readme.md

jpodwys:
Adding `superagent-cache-plugin` to the Plugins section.


pornel:
is the compatible with 1.x bit intentional? Does it mean it's compatible with 1 or later, or that it's not compatible with the current version? (3.x)?

jpodwys:
Perhaps I should reword it as "Tested against 1.x". It basically means that I don't guarantee it works with 2.x or 3.x, although a couple of people have told me they use it with 2.x and haven't had issues.

jpodwys:
I've been wanting to update it to explicitly work with the newer versions of superagent, but I haven't made time for it yet and no one has requested it. I was planning to cut major releases to match superagent, but with my superagent libs and all my cache modules, that's a lot of potential maintenance.

Issue #1152, opened 2017-01-10
Manually set content-type when using .attach


Glutch:
Hey!
I'm sending a request like this

```
request
    .post('http://website.com/upload')
    .attach('image', path)
    .send(payload)
    .set('content-type', 'multipart/form-data; boundary=----WebKitFormBoundaryKWf3FhTklvtPkprk')
```

However, the content-type gets overridden by superagents own value. Resulting in a res.req._headers like this

`content-type': 'multipart/form-data; boundary=--------------------------442535245526379898253153`

Is there any way i can force set the content-type manually?

Or perhaps figure out superagents content-type pre-post so i can use the same boundary in my .send(payload)

pornel:
No, there's no way. Setting it will break things.

In the browser version the boundary is set automatically by `FormData` element and we have no control over it.

Glutch:
@pornel So theres literally no way i can use the same boundary in my payload as my header?
If no, can i use some other framework to accomplish this?

pornel:
You can't use `attach()` and `send()` at the same time. You can either use your own content-type and your own payload with `.send().set()` OR you can let all of it being 100% auto-generated by the browser by using `.attach().field()`. They don't mix.

Glutch:
@pornel Okey, so theres still a chance! Can you send images with .send()? :) (if so how?)

pornel:
The hard way. Generate boundary, content-type header, and append image's binary data. You'll need to get Uint8Array from image Blob (using FileReader).

I'd rather do it the other way — why do you use your raw, custom payload? Can you send it as files or form fields instead?

Glutch:
@pornel I'm trying to upload a picture to a website (not mine), i've checked the network requests and i need to provide the image and a payload. 

The payload consists of information regarding the image, like this

```
------WebKitFormBoundaryKWf3FhTClwfaDpre
Content-Disposition: form-data; name="image"; filename="2684337397.jpg"
Content-Type: image/jpeg
```

And the content-type (WebKitFormBoundaryKWf3FhTClwfaDpre) must match. I'm not sure how to do it

pornel:
What do you use .send() for?

Usually you only need

```js
request
    .post('http://website.com/upload')
    .attach("image", image, "2684337397.jpg")
```

and nothing else. The boundary is supposed to be random, so needing it to match anything is very unusual.

Glutch:
@pornel The payload also consists of eventvalidation and some viewstate string, so it seems to be required

[...]

PR #1153, opened 2017-01-10
Warn when using unsupported mix of send and field

pornel:
Warn about gotcha in #1152


focusaurus:
I have no real opinion on this but would `console.warn` be slightly more semantically correct if we consider these "warnings"? I think ultimately this is going to cause misbehavior so they usually do represent actual errors in the application needing a code change to fix. (In which case `console.error` seems like the best choice)

pornel:
I have no strong opinion either way. I went with`console.error`, because it may become an error.

PR #1154, opened 2017-01-10
Throw when using unsupported mix of send and field

kornelski:
Harsher version of #1153. Probably a breaking change.

Related to #1152

I suggest releasing warning version first, and the throwing version sometime later


focusaurus:
OK I merged the warning version.

PR #1155, opened 2017-01-10
[WIP] add `Request.retry`

popeindustries:
I'm just opening this early sketch to see if there is interest in adding a `Request.retry` method. I'm aware of [segmentio/superagent-retry](https://github.com/segmentio/superagent-retry), but it relies on some hacky behaviour to reset state between attempts, and doesn't seem to work with newer versions of `superagent`. There would seem to be quite a bit of resetting necessary to make this work properly for errors, timeouts, and aborts, so I think it might be better done as a core method rather than a plugin.

In any case, I'd be happy to work on adding this if there is interest.

Progress:
- [x] Node: retry for errors
- [x] Node: retry for timeouts
- [x] Browser: retry for errors
- [x] Browser: retry for timeouts
- [x] Node: retry with redirects
- [x] refactor `end()` into 2 phases for "lightweight" retry?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -17,6 +17,7 @@ var RequestBase = require('./request-base');
 var isObject = require('./is-object');
 var isFunction = require('./is-function');
 var ResponseBase = require('./response-base');
+var shouldRetry = require('./should-retry');
 
 /**
  * Noop.
@@ -390,6 +391,8 @@ function Request(method, url) {
   this.url = url;
   this.header = {}; // preserves header name case
   this._header = {}; // coerces header names to lowercase
+  this._retries = 0;
+  this._maxRetries = 0;
   this.on('end', function(){
     var err = null;
     var res = null;
@@ -528,18 +531,18 @@ Request.prototype.auth = function(user, pass, options){
 };
 
 /**
-* Add query-string `val`.
-*
-* Examples:
-*
-*   request.get('/shoes')
-*     .query('size=10')
-*     .query({ color: 'blue' })
-*
-* @param {Object|String} val
-* @return {Request} for chaining
-* @api public
-*/
+ * Add query-string `val`.
+ *
+ * Examples:
+ *
+ *   request.get('/shoes')
+ *     .query('size=10')
+ *     .query({ color: 'blue' })
+ *
+ * @param {Object|String} val
+ * @return {Request} for chaining
+ * @api public
+ */
 
 Request.prototype.query = function(val){
   if ('string' != typeof val) val = serialize(val);
@@ -590,14 +593,51 @@ Request.prototype._getFormData = function(){
  */
 
 Request.prototype.callback = function(err, res){
+  // console.log(this._retries, this._maxRetries)
+  if (this._maxRetries && this._retries++ < this._maxRetries && shouldRetry(err, res)) {
+    var timeout = 0;
+    var responseTimeout = 0;
+    if (err && err.timeout) {
+      if (err.message.indexOf('Response') == 0) {
+        responseTimeout = err.timeout;
+      } else {
+        timeout = err.timeout;
+      }
+    }
+    return this._retry(timeout, responseTimeout);
+  }
+
   var fn = this._callback;
   this.clearTimeout();
 
   if (err) {
+    if (this._maxRetries) err.attempts = this._retries;
     this.emit('error', err);
   }
 
   fn(err, res);
+};
+
+/**
+ * Retry request
+ *
+ * @return {Request} for chaining
+ * @api private
+ */
+
+Request.prototype._retry = function(timeout, responseTimeout) {
+  this.clearTimeout();
+
+  this._timeout = timeout;
+  this._responseTimeout = responseTimeout;
+  this._aborted = false;
+  this.timedout = false;
+  delete this._timer;
+  delete this._responseTimeoutTimer;
+
+  this._endCalled = false;
+  this.end(this._callback);
+  return this;
 };
 
 /**
--- a/lib/client.js
+++ b/lib/client.js
@@ -600,7 +600,7 @@ Request.prototype.callback = function(err, res){
   this.clearTimeout();
 
   if (err) {
-    if (this._maxRetries) err.attempts = this._retries;
+    if (this._maxRetries) err.retries = this._retries - 1;
     this.emit('error', err);
   }
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -684,10 +684,6 @@ Request.prototype._isHost = function _isHost(obj) {
  */
 
 Request.prototype.end = function(fn){
-  var self = this;
-  var xhr = this.xhr = request.getXHR();
-  var data = this._formData || this._data;
-
   if (this._endCalled) {
     console.warn("Warning: .end() was called twice. This is not supported in superagent");
   }
@@ -695,6 +691,19 @@ Request.prototype.end = function(fn){
 
   // store callback
   this._callback = fn || noop;
+
+  // querystring
+  this._appendQueryString();
+
+  return this._end();
+};
+
+Request.prototype._end = function() {
+  var self = this;
+  var xhr = this.xhr = request.getXHR();
+  var data = this._formData || this._data;
+
+  this._setTimeouts();
 
   // state change
   xhr.onreadystatechange = function(){
@@ -738,11 +747,6 @@ Request.prototype.end = function(fn){
       // https://connect.microsoft.com/IE/feedback/details/837245/xmlhttprequest-upload-throws-invalid-argument-when-used-from-web-worker-context
     }
   }
-
-  // querystring
-  this._appendQueryString();
-
-  this._setTimeouts();
 
   // initiate request
   try {
--- a/lib/client.js
+++ b/lib/client.js
@@ -619,28 +619,6 @@ Request.prototype.callback = function(err, res){
 };
 
 /**
- * Retry request
- *
- * @return {Request} for chaining
- * @api private
- */
-
-Request.prototype._retry = function(timeout, responseTimeout) {
-  this.clearTimeout();
-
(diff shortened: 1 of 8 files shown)


pornel:
I think that would be a useful addition.

As you've noted, superagent has a quite a bit of internal state. Do you have a plan how to deal with that nicely?

popeindustries:
"Nicely" will definitely be the challenge ;)

I don't yet have a full picture, but it looks like `callback()` would be the best place to trigger a reset (`_aborted`, `timedout`, `called`, `_endCalled`, etc) and re-execution of `end()`. I think the real challenge will be dealing with the event handlers for the internal `req` instance. On timeout, for instance, after `abort()` is called and an `err` passed to `callback`, if we reset `_aborted`, the error handler [here](https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L597) will not return early and we'd end up with a double callback.

pornel:
I suspect it may be necessary to change `.end()` and split it into 2 phases: things done once only (e..g append query string), and things done to launch the request (the latter call being repeated when retrying).

popeindustries:
Ok, thanks. I'll start with some tests and start poking around.

popeindustries:
@pornel I think things are looking pretty good so far. I just have a couple questions if you have time to look things over:

- do you have any suggestions on how to avoid [this](https://github.com/popeindustries/superagent/blob/67c970fd78a4980f9caf5adf36433c03b6bd1075/lib/node/index.js#L641-L649) mess when resetting `_timeout` or `_responseTimeout` after they have been cleared in `clearTimeout`?
- is it necessary to reset headers [here](https://github.com/popeindustries/superagent/blob/67c970fd78a4980f9caf5adf36433c03b6bd1075/lib/request-base.js#L160)?

I've so far avoided touching `end()`. I can see how it would be more efficient to avoid initial parsing on each retry, but I'm a little hesitant to mess with anything in there if it's not absolutely necessary.

lib/client.js
```diff
@@ -590,10 +593,25 @@ Request.prototype._getFormData = function(){
  */
 
 Request.prototype.callback = function(err, res){
+  // console.log(this._retries, this._maxRetries)
+  if (this._maxRetries && this._retries++ < this._maxRetries && shouldRetry(err, res)) {
+    var timeout = 0;
+    var responseTimeout = 0;
+    if (err && err.timeout) {
+      if (err.message.indexOf('Response') == 0) {
```
pornel: That looks odd.

1. You shouldn't need to read the timeout from the error. Read request's properties set by the `timeout()` function directly
2. *if* you needed to identify error thrown by superagent, add a property to it, don't touch the message

lib/client.js
```diff
@@ -590,10 +593,25 @@ Request.prototype._getFormData = function(){
...
+      }
+    }
+    return this._retry(timeout, responseTimeout);
+  }
+
   var fn = this._callback;
   this.clearTimeout();
 
   if (err) {
+    if (this._maxRetries) err.attempts = this._retries;
```
pornel: Maybe `err.retries`?

lib/node/index.js
```diff
@@ -595,6 +598,8 @@ Request.prototype.request = function(){
     // because node will emit a faux-error "socket hang up"
     // when request is aborted before a connection is made
     if (self._aborted) return;
+    // flag retry here for out timeouts (same as above)
+    if (self._retries) return;
```
pornel: I don't understand why it skips errors here. 

Also note that all these handlers are `.once()`, so they should have been re-attached again on retry. That's why I've mentioned earlier that .end() may need to be split up into once-ever and once-per-retry functions.

lib/request-base.js
```diff
@@ -130,6 +130,50 @@ RequestBase.prototype.timeout = function timeout(options){
...
+ * Failed requests will be retried 'count' times if timeout or err.code >= 500.
+ *
+ * @param {Number} count
+ * @return {Request} for chaining
+ * @api public
+ */
+
+RequestBase.prototype.retry = function retry(count){
+  this._maxRetries = count || 1;
+  this._retries = 0;
```
pornel: That is set in `end()`, so you shouldn't need to change that

lib/request-base.js
```diff
@@ -130,6 +130,50 @@ RequestBase.prototype.timeout = function timeout(options){
...
+
+  this._timeout = timeout;
+  this._responseTimeout = responseTimeout;
+  this._aborted = false;
+  this.timedout = false;
+  delete this._timer;
+  delete this._responseTimeoutTimer;
+
+  this._endCalled = false;
+  this.end(this._callback);
```
pornel: Doesn't this break things? e.g. append query string twice?

lib/should-retry.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @param {Error} err
+ * @param {Response} [res]
+ * @returns {Boolean}
+ */
+module.exports = function shouldRetry(err, res) {
+  if (err && err.code && ~ERROR_CODES.indexOf(err.code)) return true;
+  if (res && res.status && res.status >= 500) return true;
+  // Superagent timeout
+  if (err && 'timeout' in err && err.code == 'ECONNABORTED') return true;
```
pornel: Please make sure that `.abort()` called by the user stops all retries. This may require creating a private abort for superagent's timeouts.

test/support/server.js
```diff
@@ -417,4 +417,27 @@ app.get('/if-mod', function(req, res){
   }
 });
 
+var errAttempts = 0;
+app.get('/error/ok', function(req, res) {
```
pornel: In CI we have multiple browsers running the test at the same time, and I think they're sharing the server, so they could interfere with each other's attempts.

It may be more robust to take a random ID from the client as an argument, and count retries per that ID.

pornel:
Answering your questions:

1. If you split and reuse part of `end()` that sets these timeouts, you won't need it

2. Doesn't seem needed. Does it break any tests?

3. I'd use `retries`, but `attempts` is fine too.

pornel:
> I'm a little hesitant to mess with anything in there if it's not absolutely necessary.

I could try to do that refactoring first if you want

lib/client.js
popeindustries: Yeah, this is really ugly because `_timeout` and `_responseTimeout` are zeroed out in `clearTimeout`, before `callback` is called. Any reason we can't just delete [L45-L46](https://github.com/visionmedia/superagent/blob/master/lib/request-base.js#L45-L46) in `request-base`?

pornel: Oh, yes, delete it there.

lib/node/index.js
popeindustries: This is also a little strange. After aborting `req` (timeout -> `abort()`), I delete the **old** `req`, but if I also `req.removeAllListeners()`, an unhandled "socket hang up" error is thrown from the **old** instance. If I don't remove listeners on the **old** `req` instance, then I have to skip somehow. 

I think the best approach would be to memoize the current retry index and return early if they are not the same.

lib/request-base.js
popeindustries: I'm not sure I follow. This is the main api for enabling `retry` (like `timeout`). I tried to follow the same logic as for redirects.

lib/should-retry.js
popeindustries: Good point. I'll write a test for this

test/support/server.js
popeindustries: Of course! I was wondering why the tests worked locally but not remotely.

lib/request-base.js
pornel: Sorry, I've meant it's set in https://github.com/visionmedia/superagent/pull/1155/files#diff-c24ce7e3da4c0e4ff811a2b6a76f8bd9R145

It's not an error, but it's redundant. `_retries` are never increased before `.retry()` is called, so there's no point zeroing something that's zero anyway.

lib/should-retry.js
popeindustries: I can confirm that this works fine (via test). Calling `abort()` never triggers `callback()`, so no more retries are attempted.

lib/request-base.js
popeindustries: Yes, I can see how this would be a problem. I'll take a look at refactoring `end()` next.

popeindustries:
I've split `end()` into public/private methods as you suggested. Seems to work fine. I'm not sure if there is anything else I've forgotten, but the browser tests are still failing. 

I've added an (pseudo) unique id to the requests to work around the parallelisation problem you raised, but there still seems to be something wrong with the tests that require different responses from the same endpoint.

lib/request-base.js
```diff
@@ -155,17 +155,16 @@ RequestBase.prototype._retry = function() {
...
   // node
   if (this.req) {
-    delete this.req;
+    this.req = null;
+    this.req = this.request();
   }
 
   this._aborted = false;
   this.timedout = false;
   delete this._timer;
```
pornel: I don't understand why these fields are deleted here. The timers are probably cancelled when the error happens (they should…), and should be irrelevant later.

lib/request-base.js
```diff
@@ -155,17 +155,16 @@ RequestBase.prototype._retry = function() {
...
   delete this._responseTimeoutTimer;
 
-  this._endCalled = false;
-  this.end(this._callback);
-  return this;
+  return this._end();
```
pornel: That looks good!

lib/request-base.js
popeindustries: In `_setTimeouts()` there is the following check, so if they are not removed, the new timers for retry can't be set:
```js
if (this._timeout && !this._timer) {
```
Probably better to delete or zero out in `clearTimeout()`?

pornel:
Thank you. 

The failures were only due to short timeouts in the test.

popeindustries:
That's great! Thanks for your help and patience

PR #1159, opened 2017-01-17
Add docs for HTTPS request

jwgoh:
Found out that SuperAgent supported these methods from the [test documentation](http://visionmedia.github.io/superagent/docs/test.html) (which looks a bit outdated on the GitHub pages).

Thanks for SuperAgent, I use it (almost) everyday!


pornel:
Thank you

PR #1167, opened 2017-01-28
Allow optional body/data in DELETE requests

alphashuro:
endpoints with DELETE methods can accept a body in the request, we need this feature in superagent so we can use `supertest` to test delete endpoints.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -847,16 +847,19 @@ request.options = function(url, data, fn){
 };
 
 /**
- * DELETE `url` with optional callback `fn(res)`.
+ * DELETE `url` with optional `data` and callback `fn(res)`.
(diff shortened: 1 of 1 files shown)


pornel:
The HTTP RFC says http://httpwg.org/specs/rfc7231.html#DELETE

> A payload within a DELETE request message has no defined semantics; 

so I think it's not appropriate to support it.

alphashuro:
But it is supported, and up to the server to define the semantics of the payload, so what is the harm in supporting it in superagent?

pornel:
OK, the spec says the same for GET and we already support that as a precedent.

alphashuro:
Thank you

PR #1168, opened 2017-01-28
validate retry() `count` and allow setting to 0

popeindustries:
I’m afraid I made a design error by disallowing setting retry count to `0`. This fix makes the following changes:

- passing `0/null/undefined/negative numbers` does not enable retry
- passing nothing or `true` defaults to `1` retry

### DIFF
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -140,7 +140,8 @@ RequestBase.prototype.timeout = function timeout(options){
  */
 
 RequestBase.prototype.retry = function retry(count){
-  this._maxRetries = count || 1;
+  if (count == null || count < 0) count = 1;
+  this._maxRetries = count;
   this._retries = 0;
   return this;
 };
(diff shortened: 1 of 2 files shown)


pornel:
Thanks for the PR. Validating and setting to 0 is generally a good idea.

However, interpreting negative values as 1 doesn't seem right to me. Can you change it so that:

- `retry()` == `retry(1)`
- `retry(number > 0)` sets retries
- anything else sets 0

popeindustries:
A little funky to tell the difference between `retry()` and `retry(undefined)`, but I think this should do it:

```js
  if (!('0' in arguments) || count === true) count = 1;
  if (count == null || count <= 0) count = 0;
```

pornel:
I think `arguments.length == 0` is a more common pattern.

pornel:
I've tweaked that and made a release. Thank you!

PR #1174, opened 2017-02-09
conditional withCredentials

andytwoods:
I'm sending data to different backends and found I need to be able to set withCredentials. My alternative is to duplicate the entire call to superagent. Fingers crossed.


pornel:
Yes, that was much needed. Thanks!

PR #1175, opened 2017-02-12
Fixed setting correct cookies for redirects

damoclark:
When using superagent, I discovered that if a response included a setcookie header, and was a redirect (e.g. 303), the new cookies were not being included in the redirected request, when they should be.

By applying cookies to redirect req after setting headers, rather than before, it prevents the old cookie headers overwriting the updated one from the cookiejar.

When a `cookie` header already exists in `this` which is the case with a redirect, the `// add cookies` code was being wiped out by the old cookie header.

```javascript
// add cookies
if (this.cookies) req.setHeader('Cookie', this.cookies);

for (var key in this.header) {
    req.setHeader(key, this.header[key]);
}
```
By switching them around, if an old `Cookie` header existed, it is wiped by updated cookies from the cookiejar.

```javascript
for (var key in this.header) {
    req.setHeader(key, this.header[key]);
}

// add cookies
if (this.cookies) req.setHeader('Cookie', this.cookies);
```

I'm sorry I don't have a test script for it.  It has taken me most of the day to track this down.  Thought the problem was in my code (that is where it usually is :)).

Hope you will accept my PR.

Regards,
Damien.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -613,12 +613,12 @@ Request.prototype.request = function(){
   if (url.search)
     this.query(url.search.substr(1));
 
-  // add cookies
-  if (this.cookies) req.setHeader('Cookie', this.cookies);
-
   for (var key in this.header) {
     req.setHeader(key, this.header[key]);
   }
+
+  // add cookies
+  if (this.cookies) req.setHeader('Cookie', this.cookies);
 
   try {
     this._appendQueryString(req);
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -26,6 +26,7 @@ var pkg = require('../../package.json');
 var RequestBase = require('../request-base');
 var isFunction = require('../is-function');
 var shouldRetry = require('../should-retry');
+var CookieJar = require('cookiejar');
 
 var request = exports = module.exports = function(method, url) {
   // callback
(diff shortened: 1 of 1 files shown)


pornel:

I'm worried that it might have other side effects, like replacing user's `request.set('Cookie',…)`. Maybe it should attempt to merge the header?

damoclark:
Hey @pornel 

That's a good point.  Merging the headers makes sense if a `Cookies` header already exists.  

I can change the PR to merge the headers.

But my knowledge of the complete superagent code-base is obviously lacking.  Is there code to parse stringified cookies request header back into an object that already exists in superagent?  I can't seem to find anything, but just checking.

Otherwise, I have two suggestions for your input: 
1.  Add another library dependency to superagent that does this, or
2. Use a regex to parse myself

I'm thinking option 1 would be better, but are you guys happy with adding another dependency to the project?

I did a search on npmjs and found [strict-cookie-parser](https://www.npmjs.com/package/strict-cookie-parser) which looks like it will fit the bill nicely.  It doesn't have many downloads, but then I'm thinking this is an uncommon task.

Would you like me to proceed with option 1 with this library?  Or is there a smarter way to go about it?

By the way, thanks to you and your team for superagent. :1st_place_medal: 

D.

damoclark:
Hi again @pornel 

Nevermind about my last comment.  I found a way of merging the cookies using existing libraries.

What do you think?

D.

pornel:
Instead of using the library you can concatenate both headers with `; ` as the separator. The cookie-jar is needed only to parse paths, domains, expiry dates from `Set-Cookie` header, but the `Cookie` header does not contain any of this.

damoclark:
I understand.  But the reason I used the library was to merge any duplicate cookies names, so when a new `Cookie` request header was generated from the temporary cookiejar, there would be no duplicates.  

Does it not matter if a cookie name and value pair appear twice in the `Cookie` request header?

If this doesn't matter, which will override the other?  Cookies with a matching name that comes first, or last?

D.

pornel:
You're right. This is a bit more correct.

I'll merge it when I write a test for it.

PR #1176, opened 2017-02-13
Fixed being able to define own parsers when their mime type starts with text/

damoclark:
I am attempting to set my own parser as follows:

```javascript
var jsdom = require('jsdom') ;

request.parse['text/html'] = function(res,fn) {
	res.text = '';
	res.setEncoding('utf8');
	res.on('data', function (chunk) {
		res.text += chunk;
	});
	res.on('end', function () {
		jsdom.env({
			html: res.text,
			url: url,
			done: function (err, window) { // res.body === window
				console.log('parsed document') ;
				fn(err, window);
			}
		});
	});
} ;

var agent = request.agent() ; // Gimme cookies....

agent.get(url)
.buffer(true)
.then(...) ;
```

Without this PR, the `parsed document` string is never output as the default `text` parser is used, even when the mime type returned in the response is `text/html`. 

By switching the order of the logic for determining which parser to use (the more specific test, before the less specific), any parser that is set that starts with text/ will override (as it should) the default `text` parser.

Other examples where user-supplied parsers won't execute equally apply to the following example mime types:

- text/csv
- text/rtf
- text/css

Are there any unintended consequences of this change that you can foresee?

Cheers,
Damien.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -825,11 +825,11 @@ Request.prototype._end = function() {
       } else if (isImage(mime)) {
         parser = exports.parse.image;
         buffer = true; // For backwards-compatibility buffering default is ad-hoc MIME-dependent
+      } else if (exports.parse[mime]) {
+        parser = exports.parse[mime];
       } else if ('text' == type) {
         parser = exports.parse.text;
         buffer = (buffer !== false);
-      } else if (exports.parse[mime]) {
-        parser = exports.parse[mime];
 
         // everyone wants their own white-labeled json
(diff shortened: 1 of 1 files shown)


pornel:
Thanks! 

It differs in how buffering is handled, but fortunately the rule was a bit redundant and there's a second place where it's fixed-up: https://github.com/visionmedia/superagent/pull/1176/files#diff-c24ce7e3da4c0e4ff811a2b6a76f8bd9R844

PR #1177, opened 2017-02-13
Add bearer token support

elisherer:
This would add bearer token support to the `auth` method (both for client and for node).

```js
.auth(accessToken, { type: 'bearer' })
```
Notice I made pass optional since bearer authorization doesn't need a 2nd argument.


pornel:
Thanks for the pull request.

@focusaurus what do you think about adding this?

focusaurus:
I'm OK with it, as long as people don't interpret it to mean some kind of deep OAuth 2 support.

PR #1178, opened 2017-02-14
Add optional callback to retry

amurph491:
The new retry feature is very useful, but there's no way to get any insight into which requests might be failing or why they need to be retried. I've added an optional callback parameter to `retry` so we can either report, log, or otherwise keep track of how or why retry is called for certain requests.

We're using superagent to send 100s of requests simultaneously, so knowing when we are falling into the retry logic is important. Hopefully others will find this useful as well.

### DIFF
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -139,12 +139,13 @@ RequestBase.prototype.timeout = function timeout(options){
  * @api public
  */
 
-RequestBase.prototype.retry = function retry(count){
+RequestBase.prototype.retry = function retry(count, fn){
   // Default to 1 if no count passed or true
   if (arguments.length === 0 || count === true) count = 1;
   if (count <= 0) count = 0;
   this._maxRetries = count;
   this._retries = 0;
+  this._retryCallback = fn;
   return this;
 };
 
@@ -157,7 +158,9 @@ RequestBase.prototype.retry = function retry(count){
 
 RequestBase.prototype._retry = function() {
   this.clearTimeout();
-
+  if (this._retryCallback) {
+    this._retryCallback(this);
+  }
   // node
   if (this.req) {
     this.req = null;
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -135,6 +135,7 @@ RequestBase.prototype.timeout = function timeout(options){
  * Failed requests will be retried 'count' times if timeout or err.code >= 500.
  *
  * @param {Number} count
+ * @param {Function} [fn]
  * @return {Request} for chaining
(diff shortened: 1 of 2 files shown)


pornel:
There's `err.retries`. I suppose we could add `res.retries` too. Would just the number of retries be enough for you?

amurph491:
That's a good thought - ideally, i think it would be nice to see _why_ the requests failed before retrying and potentially succeeding, but at least including the number of retries on the `res` object would allow some level of investigation/reporting on requests that are needing to be retried

pornel:
I'm not sure where to take design of this feature. 

Right now it's a bare minimum, but retries overall can get very complex. Should superagent try to handle them all? For example, you may want to retry after a delay, and the delay should probably grow exponentially or with an s-curve (unless there's an `online` event, which should interrupt the wait). You might want to retry some errors, but not others. Maybe retries caused by timeouts should increase the timeouts.

amurph491:
Totally agree - I like the idea of having the option to add some finer control over the retries for certain requests and responses. To start at least, I think it'd be sufficient to leave this retry logic up to the user, but at least provide this callback option to allow aborting a sequence of retries, introducing delays, or any other logic that may suit specific uses. I'm happy to build this into this PR, there would need to be some consensus on what the passed in callback function should return (like you mention, a promise, a superagent request, or a boolean)

pornel:
OK, so let's try this:

Sync MVP:

- the callback should receive only one argument - the error that has happened (so it can check the error code)
- if the callback returns exactly boolean `false`, then the request isn't retried and fails with the error
- if the callback throws, the request fails with the thrown error
- if the callback returns `true` or `undefined` then continue as usual

And an extra feature:

- if the callback returns an object with a `.then` property, then it's resolved using `Promise.resolve(obj)`, and the request waits for the promise:
    - when the promise resolves successfully, interpret it as a result from the callback (so give up if it's `false`, continue otherwise)
    - when the promise rejects, then fail the request with the error from the promise

I think it could be used as:

```js
// Retry only connection errors
req.retry(err => err.code != 'ECONNREFUSED'); 

// Retry after a delay, increasing with number of failures
req.retry(err => new Promise(cb => setTimeout(cb, 1000 * err.retries)));

// Just log
req.retry(console.warn); 
```

ryanhugh:
Any update on this? I would totally use the features @pornel is talking about if they were live right now :)

abrkn:
I'm very interested in this feature. There are certain errors returned from the API I consume which mean I should retry.

peebles:
me too

PR #1180, opened 2017-02-14
More friendly for frontend-bundlers

gerhardsletten:
Since lib/client.js is only consumed by browsers, `component-emitter` can be required directly. The way it is now, bundlers will have to memorize this `emitter` shim when resolving `superagent` 

This will make superagent work in Fuse-box


pornel:
That's done on master now

PR #1189, opened 2017-03-08
Removing memory leaks

andrija-hers:
The proposed PR affects only the __Node.js__ side.

We use `superagent` implicitly, through the `image-resolver` module that we use heavily on our servers (up to 4500 `superagent` requests per hour). We found that in 10 hours around 400 MB of garbage gets produced. This figure of 400 MB is obtained by comparing a relatively steady memory usage of our server of 300 MB (with this PR applied) compared to 700 MB of memory usage after 10 hours of operation (without this PR applied); of course, without this PR, the memory consumption of our server keeps rising all the time, in a saw-tooth shaped manner.

Regarding the very PR:

1. `Request` had to undergo the hardest refactor - `UnitRequest` was introduced to model a single request to a remote server.
2. All other changes are rather trivial; they were introduced to let the garbage collector do its job. Greatest care was taken in order to (according to the use case) properly remove appropriate event handlers so that all bound references get released.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -58,6 +58,12 @@ exports.agent = require('./agent');
  */
 
 function noop(){};
+function doRemoveListener(eventemitter, eventname, eventhandler) {
+  eventemitter.removeListener(eventname, eventhandler);
+  if (eventemitter.listenerCount(eventname) < 1) {
+    eventemitter.on(eventname, noop);
+  }
+}
 
 /**
  * Expose `Response`.
@@ -131,6 +137,7 @@ function _initHeaders(req) {
  * @api public
  */
 
+
 function Request(method, url) {
   Stream.call(this);
   if ('string' != typeof url) url = format(url);
@@ -147,7 +154,8 @@ function Request(method, url) {
   this.qsRaw = [];
   this._redirectList = [];
   this._streamRequest = false;
-  this.once('end', this.clearTimeout.bind(this));
+  this._unitRequest = null;
+  //this.once('end', this.clearTimeout.bind(this));
 }
 
 /**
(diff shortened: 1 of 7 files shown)


focusaurus:
Thanks for this great effort! I'll watch for additional commits addressing the unit test failures.

andrija-hers:
@focusaurus , sorry for this, I wouldn't have sent the PR if I had recognized these failures. Obviously, I was unaware of how to run tests - simple `mocha` call succeeded on my side (node 6.5.0, 165 passing in 17s). If that's not a problem for you, when you find time, please let me know about the steps needed to run _all_ the tests.

andrija-hers:
`npm test`, silly me...
We'll be back with all the issues covered :)

pornel:
Are you sure that *all* of it is required? Did you find the root cause of the increased memory usage? I would prefer not to do a major refactoring of the whole library on a hunch.

I'm especially concerned about changes to closures. Avoidance of cycles caused by closures looks like the workaround that used to be needed for IE6, but it was a problem very specific to IE's DOM, and it doesn't apply to V8.

I see a potential for unbounded memory growth in `image-resolver`:
https://github.com/mauricesvay/ImageResolver/blob/023ee8944ec4ca69932781f2282c53b38c6d9665/src/Utils.js#L33,L35

so I think that's where your leak is, not in superagent. Thinning of superagent's objects just makes the stored-forever `response` objects smaller.

andrija-hers:
@pornel , many thanks for spending your time on my case :)
However, `image-resolver` is already taken care of (by me again :| ), and that particular line you're referring to is already covered (among several other "leaky" lines).

On the other side, regarding the very garbage manipulation, we learned a very important lesson. Here's the situation:

- on our server we use a single process for the whole server
- our server does a lot of stuff, one of them being the image resolving
- the whole result of image resolving is a single string; but, in order to produce a string that will be stored in memory, up to `1e4` temporary objects might be produced

The case described above stabilizes at around 300 MB.

Then we "split" the server process into 
1. the main server process
2. the image resolver child process

The important thing with this setup is that the child process stores _nothing_, just reports the resolved string back to the parent process.

We ended up in the parent process taking up <70MB and the child process swinging wildly in the (60, 250) MB range, but for most of the time taking under 100MB.

The reason is that frequently a small number of long-living objects might find themself in a memory chunk full of garbage. These useful objects keep the whole chunk from being free-d.
Moral of the story: whenever possible, "outsource" a "pollution-intensive" job to a separate process that will produce no long-living objects.

Finally, our tests have shown that `superagent` is working perfectly with no interventions, and the `image-resolver` _really_ needed memleak cleaning + additional options for avoiding the cache-ing and utilizing the `timeout` feature of `Response`.

Guys, many thanks for the great work on `superagent`! There will be no PRs from our side regarding memory handling :)

PR #1190, opened 2017-03-11
Check for ownProperty before adding header

vieiralucas:
Hello everyone

An issue was filled on chai-http chaijs/chai-http#153

The issuer was adding a property to the prototype of Object which was making `superagent` throw a TypeError when setting a header. (I know you should not be adding properties to the prototype of Object, but the truth is that you can).

So, to prevent that for happening. This PR adds a check for `hasOwnProperty` before setting the header.

I also added a test that reproduces it.

Thank you 😄


pornel:
Thank you

PR #1192, opened 2017-03-17
Read responseType property from the correct object

jdupouy:
The responseType property wasn't read from the correct object so it wasn't taken into account.


pornel:
Good catch, thank you.

Issue #1194, opened 2017-03-18
Support for retrying failed CORS requests


mwolson:
When using the retry() functionality on a bad connection, the browser can sometimes fail during a CORS OPTIONS request, which currently doesn't get retried. Superagent throws an error with crossDomain: true property in this case. We should retry instead, since the next attempt might succeed.

mwolson:
Fixed in acaf84d

PR #1195, opened 2017-03-18
Allow crossDomain errors to be retried (#1194)

mwolson:
When using the retry() functionality on a bad connection, the browser can sometimes fail during a CORS OPTIONS request, which currently doesn't get retried. Superagent throws an error with crossDomain: true property in this case. We should retry instead, since the next attempt might succeed.


pornel:
Thanks.

That's a good idea. True CORS failure will be blocked quickly, so there's no harm retrying.

mwolson:
Added some docs for `.retry()` as well, since I noticed they were missing.

pornel:
Thank you

PR #1199, opened 2017-03-29
remove duplicated query params

bwidtmann:
this bug occurs when using new retry feature and using query params in the URL string instead of passing it via the .query({qs: 'present'}) object.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -306,7 +306,10 @@ Request.prototype.accept = function(type){
 
 Request.prototype.query = function(val){
   if ('string' == typeof val) {
-    this.qsRaw.push(val);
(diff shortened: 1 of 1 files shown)


pornel:
Do you have an example of a bug that happens with retry?

This is supported:

```js
request.get().query('arr[]=1').query('arr[]=1')
```

So this is a breaking change, and the fix must be different.

bwidtmann:
`request.get('http://www.google.com?bla=1&blup=1).retry(1)`

the first request is ok, the second request would be:

`http://www.google.com?bla=1&blup=1&bla=1&blup=1`

you know what I mean?

pornel:
Thanks. I've made a fix for the root cause in #1200

PR #1207, opened 2017-04-11
Add options passpharse and rejectUnauthorized for tls with self-signe…

talaevlev:
…d certificate

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -507,6 +507,32 @@ Request.prototype.ca = function(cert){
 };
 
 /**
+ * Set the passphrase authority option for https request.
+ *
+ * @param {string} passphrase
+ * @return {Request} for chaining
+ * @api public
+ */
+
+Request.prototype.passphrase = function(passphrase){
+    this._passphrase = passphrase;
+    return this;
+};
+
+/**
+ * Set the rejectUnauthorized option using self-signed sertificate for https request.
+ *
+ * @param {boolean} rejectUnauthorized
+ * @return {Request} for chaining
+ * @api public
+ */
+
+Request.prototype.rejectUnauthorized = function(rejectUnauthorized){
+    this._rejectUnauthorized = rejectUnauthorized;
+    return this;
+};
+
+/**
  * Set the client certificate key option for https request.
  *
  * @param {Buffer | String} cert
@@ -596,6 +622,8 @@ Request.prototype.request = function(){
   options.key = this._key;
   options.pfx = this._pfx;
   options.cert = this._cert;
+  options.rejectUnauthorized = this._rejectUnauthorized;
+  options.passphrase = this._passphrase;
   options.agent = this._agent;
 
   // initiate request


pornel:
No, sorry. That option disables all security of HTTPS, and it's poorly named for such a dangerous option.

talaevlev:
And what should I do?
I have only self-signed certificate for tests.

pornel:
My concern here is that if there's an option in code that disables all security, then it's likely to be accidentally committed and left enabled in production.

If your test server is visible on the internet, then you can get a real cert from certbot/let's encrypt for free. 

If your test server is `localhost` or some other private, then I suggest making the server URL configurable and use HTTP URL in dev (and a domain that isn't accessible from production servers, so erroneous use of dev URL in production will be noticed quickly).

And finally, you can create your own CA and sign your cert with it, and set the CA with `.ca()` option, although it is a bit of a hassle.

Bluscream:
I would like to see such a option too. Some people route their SSL traffic to software like Adguard which uses it's own cert validation so node does not need to take care of that.

rusekr:
@pornel  You are rejecting configurability. Let people decide themself how to use this basic options. It's basic nodejs functionality. I mentioned here https://github.com/visionmedia/superagent/issues/188#issuecomment-336725716 real life use of this option.

There are not always YOUR systems with bad certificates in real life and you must integrate with them anyway (responsibility for potential mitm - another question). 

p.s. Meanwhile you can make checking cert not only required but require to save and check it's fingerprint because now MITM can be made with green certificates which any breacher can get from mozilla freely. And no default rejectUnauthorized: true can say the differece. Security is configured by means, not for all.

pornel:
@rusekr Mozilla gives certificates for freely as in "$0", but not freely as in "any cert to anyone", so they still maintain strong chain of trust that self-signed/unverified certificates don't.

rusekr:
This company staff uses only paid certificates. Its decision of big management.

About strong chain i wrote in p.s. - if i can hack your dns - mozilla gives me plenty space to have green browser bar for my fake servier for you without you notice (except fingerprint saving/checking enabled on your client). Mobile browsers dont even allow you to see which cert is actually on server you connected to. (;

pornel:
All DV certs are vulnerable to hijacking of authoritative DNS (i.e. not user's Wi-Fi, but networks at ISPs/hosting), regardless whether you pay for them or not. 

Let's Encrypt is on par or more secure than other CAs thanks to using ACME challenges (rather than e.g. SMTP), [and they are adding even stronger protections](https://community.letsencrypt.org/t/validating-challenges-from-multiple-network-vantage-points/40955) against BGP-based attacks. (not that it matters given overall design of CA PKI, but I'm just correcting a rumor you're spreading).

There's Certificate Transparency which is another method of detecting successful impersonations.

And if by "hacking my DNS" you mean DNS queries made by my browser, then this is exactly what verification of chain of trust protects against. So no, you definitely can't get a green bar this way (unless I foolishly made my browser behave like `rejectUnauthorized=false`).

rusekr:
>> verification of chain of trust protects against
Verification chain is for the convenience of certificate authorities. No need for them to revoke root cert if just one of its children (or child of children) compromised. It's simple delegation of trust.

 If i spoof your dns requests your browser will be checking my certificate chain, not yours (;

meanwhile nodejs is not a browser. it's server to server communication and i can check even fingerprint of my expired certificate and it will be proof that it is not switched with fake.

And its went to nowhere. Got it. Your locked freedom in your lib for no reason.

pornel:
No, you're technically incorrect about this.

rusekr:
You are using ssh - its same as non-browser to server ssl. There is no green-mafia in it for all good. Here only known or unknown key fingerprints as i want with rejectUnauthorized: true in that case.

pornel:
No, SSH's TOFU model is different than CA PKI and has different security trade-offs.

> If i spoof your dns requests your browser will be checking my certificate chain, not yours (;

And it will check it and find it's not authorized by any CA, and reject it. That's exactly what needs to happen, yet you still seem to be advocating for removal of this check that is critical for security.

I'm locking this topic so you have time to cool down and not flood everyone in this thread with notifications.

Issue #534, opened 2015-01-22
.send becomes .query for GET requests, but will error out when used for HEAD requests


mikemaccana:
From the docs:

> When issuing a GET request the res.send(obj) method will invoke res.query(obj),

For consistency's sake, the same should apply to HEAD requests, but doesn't. When using `.send` rather than `.query` on a HEAD request superagent will error:

```
http.js:851
    throw new TypeError('first argument must be a string or Buffer');
          ^
TypeError: first argument must be a string or Buffer
    at ClientRequest.OutgoingMessage.write (http.js:851:11)
    at ClientRequest.OutgoingMessage.end (http.js:981:16)
    at Request.end (/Users/someuser/Documents/someproject/node_modules/superagent/lib/node/index.js:968:9)
    at Object.isDuplicate (/Users/someuser/Documents/someproject/node_modules/mycognition-usermanagement/index.js:44:5)
    at Object.<anonymous> (/Users/someuser/Documents/someproject/node_modules/mycognition-usermanagement/example.js:3:27)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
```

englercj:
I also don't think this is true for the browser version, which seems to operate differently. I find I have to actually call `query` instead of `send` for the browser version.

acontreras89:
What's the status on this? 

Perhaps [this line of code](https://github.com/visionmedia/superagent/blob/master/lib/client.js#L810) should be changed to:

```js
if (data) req.query(data); // rather than req.send
```

As described by [the HTTP spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4) (credits to [this SO answer](http://stackoverflow.com/questions/4529066/post-data-with-head-request#answer-4529097)):

> The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.

It would be nice to have both functions (`.get` and `.head`) behave the same way.

focusaurus:
God I wish our APIs would just consistently do the same thing instead of overloading generic methods like `send` and trying to be clever. That said, if we behave that way and document it for GET, I'd agree to at least make it consistent for HEAD, although ultimately I'd rather not have a method that does 2 totally different things depending on the state of the request built up in the method chaining DSL so far.

focusaurus:
[...]

PR #1212, opened 2017-04-27
Use query instead of send for HEAD requests

acontreras89:
As per [my comment](https://github.com/visionmedia/superagent/issues/534#issuecomment-297592130) on issue #534.

I don't know if you'll be willing to accept such change, but it didn't require much effort.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -807,7 +807,7 @@ request.get = function(url, data, fn){
 request.head = function(url, data, fn){
   var req = request('HEAD', url);
   if ('function' == typeof data) fn = data, data = null;
-  if (data) req.send(data);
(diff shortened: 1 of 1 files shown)


focusaurus:
Should we make the corresponding change to [the node side](https://github.com/focusaurus/superagent/blob/master/lib/node/index.js#L931) as well?

acontreras89:
@focusaurus I'm not familiar with the node implementation of this. I took a look but it seems it's already treating GET and HEAD requests the same way, even though they are not treated as in the browser version.

kornelski:
We've done this in ce2f27d5

acontreras89:
:clap:

PR #1226, opened 2017-05-15
Update Readme.md

jpodwys:
superagent-cache is now compatible with superagent `1.x`, `2.x`, and `3.x`.


focusaurus:
Thanks!

focusaurus:
FYI my tests were passing locally but this failed in Travis. I wiped  `node_modules` and re-installed clean and I can reproduce the test failure, which is unrelated to this PR code change but probably due to some dependency changing behavior in our test web app. Investigating...

PR #1227, opened 2017-05-15
Fix xml tests that broke with mime update

focusaurus:


focusaurus:
Looks like mime 1.3.4 had `xml` as `application/xml` and by 1.3.6 it switched to `text/xml`. I don't know the backstory behind that nor am I sure of what the semver/breakage implication is for end users of superagent.

focusaurus:
@pornel Thoughts on this? 

I confirmed it was mime 1.3.6 that introduced this change.

```sh
/tmp/20170524-100433 -> npm i mime@1.3.5
mime@1.3.5 node_modules/mime
/tmp/20170524-100433 -> node
> require('mime').types.xml
'application/xml'

/tmp/20170524-100433 -> npm i mime@1.3.6
mime@1.3.6 node_modules/mime
/tmp/20170524-100433 -> node
> require('mime').types.xml
'text/xml'
```

I did some github research trying to understand the back story for this change but the rabbit hole got the best of me.

focusaurus:
Also FYI I asked for help here: https://github.com/broofa/node-mime/issues/158

I see no mention of this change in the mime-db history file: https://github.com/jshttp/mime-db/blob/master/HISTORY.md

I'm a bit perplexed at this point. Hopefully someone closer to the issue might shed some light so we can merge this and get the tests passing again.

PR #1228, opened 2017-05-17
Send payload in query string for GET and HEAD shorthand API

focusaurus:
Partially addresses #534. Browser and node are now consistent for
both GET and HEAD, but only the shorthand APIs behave this way.


focusaurus:
Note 2 mime xml test failures are expected. There's a separate open PR that fixes those. All other tests should be passing for both node and the browser.

PR #1229, opened 2017-05-17
Docs to master branch

focusaurus:
Github pages now supports serving the static site from the master branch `docs` directory. The `gh-pages` branch was confusing because the only things we used from there were the docs but there was also a ton of ancient code in there. With this change, we should be able to go into the github settings and configure our github pages site to be based off master branch `docs/index.html`. Then we can delete the `gh-pages` branch which should simplify things.

I also switched our table of contents generator to an actively-maintained, 0-dependency alternative.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,20 +36,20 @@ superagent.js: lib/node/*.js lib/node/parsers/*.js
 test-server:
 	@node test/server
 
-docs: index.html test-docs docs/index.md
+docs: docs/index.html test-docs docs/index.md
 
-index.html: docs/index.md docs/head.html docs/tail.html
+docs/index.html: docs/index.md docs/head.html docs/tail.html
 	marked < $< \
 		| cat docs/head.html - docs/tail.html \
 		> $@
 
 docclean:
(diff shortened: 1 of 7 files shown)


pornel:
But that wasn't the reason for gh-pages branch. The reason was it contains large auto-generated files that cause repo bloat and merge conflicts.

pornel:
No, because it's OK to force-push a fresh version gh-pages branch and erase its history, which removes the bloat. But we can't do that with the master branch, as that'd break all hashes/tags/etc.

pornel:
The test.html is a bit of a mess. I don't know, maybe someone is using it? It does provide a lot of code, so maybe it comes up in searches sometimes?

focusaurus:
OK I'll open a new PR targeting gh-pages.

pornel:
Ah, sorry for confusion. They way I currently work with it is:

- changes to documentation sources (index.md, Makefile) are in master
- the branch with result of `make docs` is pushed to gh-pages

so in that way gh-pages is disposable.  I've copied your PR to master branch.

Issue #1213, opened 2017-04-29
post pfx has problem


runhwguo:
when i use superagent post,  .pfx(),  it always goes wrong

Error: error:0906D06C:PEM routines:PEM_read_bio:no start line

my code is here:
  // let response = await superagent
  //   .post(URL_REFUND)
  //   .send(formData)
  //   .pfx(PFX)
  //   .key(MCH_ID)
  //   .charset(config.common.char_set_utf8);

but when i use request library, it is ok!

pornel:
Unfortunately I don't know why this is happening. We're just passing the argument straight to [Node's request](https://nodejs.org/api/https.html):

https://github.com/visionmedia/superagent/blob/8f1a90cbcc9aac066bfe58b4ad38ff2b3092bb46/lib/node/agent.js#L79

so if it works with Node, it should work with Superagent too.

PR #1230, opened 2017-05-18
Bugfix/pfx passphrase

PWesterdale:
Node's internal method for passing a PFX passphrase has changed, This brings support in superagent up to date.


pornel:
Great. Thank you. That probably fixes #1213

kornelski:
It looks like this has stopped working again. 

https://travis-ci.org/visionmedia/superagent/jobs/331565966#L2282

Does it work for you? Could you take a look?

PR #1240, opened 2017-06-23
support TCP_NODELAY option

xiamengyu:
Add setNoDelay method for Request, this method can enable `TCP_NODELAY` option.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -622,9 +622,8 @@ Request.prototype.request = function(){
   var req = this.req = mod.request(options);
 
   // set tcp no delay
-  if (true === this._noDelay) {
-    req.setNoDelay(true);
-  }
+  req.setNoDelay && req.setNoDelay(!!this._noDelay);
+
 
   if ('HEAD' != options.method) {
     req.setHeader('Accept-Encoding', 'gzip, deflate');
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -622,8 +622,9 @@ Request.prototype.request = function(){
   var req = this.req = mod.request(options);
 
   // set tcp no delay
-  req.setNoDelay && req.setNoDelay(!!this._noDelay);
-
+  if (this._noDelay === true && req.setNoDelay) {
+    req.setNoDelay(true);
+  }
 
   if ('HEAD' != options.method) {
     req.setHeader('Accept-Encoding', 'gzip, deflate');
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -551,6 +551,17 @@ Request.prototype.cert = function(cert){
 };
 
 /**
+ * Set tcp no delay option
+ *
+ * @param noDelay
+ * @returns {*}
+ */
+Request.prototype.setNoDelay = function(noDelay){
+  this._noDelay = noDelay === true;
+  return this;
+};
+
+/**
  * Return an http[s] request.
  *
  * @return {OutgoingMessage}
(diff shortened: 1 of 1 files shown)


pornel:
I don't see a reason to add buffering delay for requests. Why do you need this?

xiamengyu:
This is not add a delay for a request, this is support TCP NO_DELAY option to disable TCP Nagle's algorithm https://en.wikipedia.org/wiki/Nagle%27s_algorithm.
We found some TCP packages are delayed  because of TCP Nagle's algorithm.
This commit can support under layer TCP NO_DELAY option to disable TCP Nagle's algorithm.

pornel:
1. In Node Nagle's algorithm is off by default, so without this change it's always off (good). With this change it's on or off.

2. I don't see default set for `this._noDelay` anywhere, so it's `undefined`, and `!!this._noDelay` makes `false`, so your change makes it slow by default.

pornel:
> https://nodejs.org/api/net.html#net_socket_setnodelay_nodelay
>  noDelay defaults to true.

So if true is the default used by superagent, the only use for this function is to set it to false. Why would you need to set nodelay to false?

xiamengyu:
'noDelay defaults to true'  means the param noDelay is true when you call setNoDelay without pass a param.
The same web page says 'By default TCP connections use the Nagle algorithm'. So I need a option to disable Nagle algorithm(enable TCP NO_DELAY).

xiamengyu:
Here is the source code of socket.setNoDelay:
`Socket.prototype.setNoDelay = function(enable) {
  if (!this._handle) {
    this.once('connect',
              enable ? this.setNoDelay : () => this.setNoDelay(enable));
    return this;
  }

  // backwards compatibility: assume true when `enable` is omitted
  if (this._handle.setNoDelay)
    this._handle.setNoDelay(enable === undefined ? true : !!enable);

  return this;
};`
[lib/net.js](https://github.com/nodejs/node/blob/095c0de94d818088cacf2c33ad4913768c15024a/lib/net.js)

pornel:
Ah! I see. I'm sorry I misread that documentation page.

I've tried this change and I tried to see how much it affects the speed, but I'm not seeing any effect:

```js
describe('benchmark', function(){
    it('nodelay', function(){
      this.timeout(1000000);
      const https = require('https');
      return Array.from(Array(100)).reduce(prev => prev.then(() => {
        return request
          .head('https://geekhood.net/robots.txt') // remote serv in case localhost is special
          .agent(new https.Agent({ keepAlive: false }))
          .set('connection', 'close')
      }), Promise.resolve());
    });
  });
```

With and without nodelay I'm seeing about the same speed (tested on macOS).

xiamengyu:
Thank you for you merged this commit and added benchmark for it. 
The delay of  Nagle algorithm happens on some special situations,  that not always happen.
Here is the Algorithm:
```
if there is new data to send
  if the window size >= MSS and available data is >= MSS
    send complete MSS segment now
  else
    if there is unconfirmed data still in the pipe
      enqueue data in the buffer until an acknowledge is received
    else
      send data immediately
    end if
  end if
end if
```
And if your OS enable `TCP_NODELAY` already, this will has no extra  effect.
By the way, may be it's a good choice for user (who use **superagent**) to choose enable `TCP_NODELAY` or not.
I see `TCP_NODELAY` is enabled by default in **superagent** now, and user can not disable it. 
Nagle's algorithm sometimes can save network resource, may be some situations need Nagle's algorithm, but user can not disable `TCP_NODELAY` in **superagent**.

pornel:
Yes, since I didn't see any negative side effect, I've enabled it all the time just in case.

As far as I understand cost/benefit of Nagle's algorithm depends on how the data is sent — whether it's sent all at once, or sent bit by bit.

In case of superagent user has absolutely no control over this. All of the sending happens in Node's underlying HTTP request library. So I don't see why user would need to have control over this, since it doesn't depend on the user, but depends on Node's implementation. In case of superagent there should clearly exist a right and wrong setting, and there shouldn't be a need to expose to user ability to set it to a wrong value.

PR #1248, opened 2017-07-11
Update supported node version to >= 4.0

jwgoh:
https://github.com/visionmedia/superagent/blob/master/History.md#300-2016-11-19

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,6 @@
   },
   "main": "./lib/node/index.js",
   "engines": {
-    "node": ">= 0.12"
+    "node": ">= 4.0"
   }
 }


caicai0:
I used it for loading.
var loadingUrl = [];
function load(url){
loadingUrl.push(url);
    superagent.get(url).end(function(err,res){
        loadingUrl.remove(url);
       //other code
    });
}

when I used this load() function too many times for diffrent urls.I find that there are away some urls in loadingUrl.I do not know why

pornel:
@caicai0 Try using constructs like `Promise.all()` or use `async/await`.

However, please do not post in these comments, as they are about a pull request, which is totally unrelated to your problem.

PR #1254, opened 2017-07-26
enable buffer default for node

lyykfi:


pornel:
Unfortunately this is a breaking change

niftylettuce:
What was the purpose of this @lyykfi?

PR #1255, opened 2017-08-01
Consolidate auth args handling in node/browser

### LINKED ISSUE #796: auth args handling different on node/browser APIs
The `request.auth()` code is slightly different between node and browser, and there's no technical reason why it couldn't be identical. In particular, the node code has a bit more automagic handling of permutations of `user, pass` arguments.

Would be straightforward to add some extra tests and get them to parity. Existing tests are pretty good so this would be a good first contribution.

edmundoa:
As described in #796, `auth` arguments handling was quite different in the code running on a browser and Node, which could lead to inconsistent results between both platforms.

This PR tries to bring them into an equal state, and also updates the test cases to cover the same situations in both platforms.

Fixes #796

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -518,21 +518,14 @@ Request.prototype.auth = function(user, pass, options){
     }
   }
 
-  switch (options.type) {
-    case 'basic':
-      this.set('Authorization', 'Basic ' + btoa(user + ':' + pass));
-    break;
+  var encoder = function(string) {
+    if ('function' === typeof btoa) {
+      return btoa(string);
+    }
+    throw new Error('Cannot use basic auth, btoa is not a function');
+  };
 
-    case 'auto':
-      this.username = user;
-      this.password = pass;
-    break;
-
-    case 'bearer': // usage would be .auth(accessToken, { type: 'bearer' })
-      this.set('Authorization', 'Bearer ' + user);
-    break;
-  }
-  return this;
+  return this._auth(user, pass, options, encoder);
 };
 
 /**
--- a/lib/client.js
+++ b/lib/client.js
@@ -507,8 +507,10 @@ Request.prototype.accept = function(type){
  */
 
 Request.prototype.auth = function(user, pass, options){
-  if (typeof pass === 'object' && pass !== null) { // pass is optional and can substitute for options
+  if (1 === arguments.length) pass = '';
+  if (typeof pass === 'object' && pass !== null) { // pass is optional and can be replaced with options
     options = pass;
+    pass = '';
   }
   if (!options) {
     options = {
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -485,15 +485,12 @@ Request.prototype.auth = function(user, pass, options){
   if (!options) {
     options = { type: 'basic' };
   }
-  switch (options.type) {
-    case 'bearer':
-      return this.set('Authorization', 'Bearer ' + user);
 
-    default: // 'basic'
-      if (!~user.indexOf(':')) user = user + ':';
-      var str = new Buffer(user + pass).toString('base64');
-      return this.set('Authorization', 'Basic ' + str);
-  }
+  var encoder = function(string) {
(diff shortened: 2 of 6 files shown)


pornel:
Thank you for the PR. It's very useful. The code looks good, but there are some failures in CI. Can you check it?

edmundoa:
I will take a look, thank you for your review!

edmundoa:
It looks like Safari and IE send cached passwords, when the HTTP basic auth password is empty, as it is in a couple of the tests I added on this PR. That makes the test server return a 401 challenge, and browsers show a prompt to ask for new credentials, causing the test to time out and fail.

I have been looking for a while, but I don't know how can I clear those cached credentials, any ideas or suggestions?

pornel:
This hasn't been a problem previously, so I think it's not a cause of caching/clearing, but difference in what these browsers were told to do.

edmundoa:
You didn't see the error before because there were no tests for this in the browser, they got added in this PR 7a822362148983ef6e47cfedaf0e31139dda6178.

Just to verify the issue existed before, I ran the new tests locally in Safari against the old code, and they showed the same behaviour as the CI reported. That of course doesn't mean I didn't do something wrong on the tests, but I can't see where 😄 

Yesterday I didn't have that much time to write more details on why I think the error we see is cache-related, but here they are now.

To debug the issue I added logs into the code doing the basic auth in the browser (via superagent) and the express middleware that is doing the basic auth on the test server side.

This is what I get in the browser console when running the tests in Chrome 60 (or Firefox 55):
```
basic auth with "foo", "bar"
basic auth with "foo", "bar"
basic auth with "foo", "bar"
basic auth with "foo", ""
basic auth with "foo", ""
```

This is what the server is logging:
```
auth attempt  user: "foo"  pass: "bar"  successful?  true
auth attempt  user: "foo"  pass: "bar"  successful?  true
auth attempt  user: "foo"  pass: "bar"  successful?  true
auth attempt  user: "foo"  pass: ""  successful?  true
auth attempt  user: "foo"  pass: ""  successful?  true
```

This is what I get in the browser console when running the tests in Safari 10.1.2 (or IE 11):
```
[Log] basic auth with "foo", "bar" (test-bundle.js, line 1307)
[Log] basic auth with "foo", "bar" (test-bundle.js, line 1307)
[Log] basic auth with "foo", "bar" (test-bundle.js, line 1307)
[Log] basic auth with "foo", "" (test-bundle.js, line 1307)
[Log] basic auth with "foo", "" (test-bundle.js, line 1307)
```

This is what the server is logging:
```
auth attempt  user: "foo"  pass: "bar"  successful?  true
auth attempt  user: "foo"  pass: "bar"  successful?  true
auth attempt  user: "foo"  pass: "bar"  successful?  true
auth attempt  user: "foo"  pass: "bar"  successful?  false
auth attempt  user: "foo"  pass: "bar"  successful?  false
auth attempt  user: "foo"  pass: ""  successful?  true
```

As you can see, in Safari/IE there are two extra attempts to authenticate with `foo` and password `bar`, even when it should use `foo` and an empty password. The browser is logging the right credentials and they also appear correctly in the network request in the browser's developer tools:

screen shot 2017-08-03 at 13 24 34

You can see that the request includes an `Authorization` header with `Zm9vOg==` as encoded credentials, which is exactly what I get when running `btoa('foo:')` in the browser console.

It is at that point when the browser displays a prompt asking for new credentials, and also when the tests time out if the prompt is not closed.

I'll try to further debug the request being sent by the browsers with Wireshark or something similar, to see if the request is actually wrong or something else is going on, but maybe someone has a better idea on how to fix this, or at least what is wrong with it.

In case it wasn't clear before, I am not saying there's nothing wrong with the code I submitted, just that the only explanation I could find so far after looking into the issue is not pointing at the changes in the PR 🙂

pornel:
Ahh, I see :)

pornel:
Sorry for the delay. I've cherry-picked your changes in 06d7865d5ab90cf0a84c914fd63e7a5e7be96b8f

PR #1256, opened 2017-08-02
Fixed typo in Readme.md

MichalCafe:
"sporting" -> "supporting"


pornel:
I think that's intended here, [sporting](https://en.wiktionary.org/wiki/sport#Verb) meaning "to have as a notable feature."

PR #1260, opened 2017-08-08
Documented FormData support in .send()

lorenzos:
I wanted to use superagent to send an HTML form, because of some `<input type="file">` fields. Browser support section in the master branch's README hints that `FormData` is somehow supported, but I couldn't find any mentions of it in the documentation website. After blindly trying using `.field()` and `.append()`, I found out that it works with `.send()`.


pornel:
Thank you

PR #1278, opened 2017-09-21
support calling .then() and .catch() after .pipe()

### LINKED ISSUE #1277: Calling .then() after .pipe() causes "write after end error"
I get that `.then()` is designed to send the request when not using `.pipe()`.

But how about making it so that after `.pipe()` is called, `.then()` just waits for the request to finish instead of trying to write again?

jedwards1211:
fix #1277


pornel:
Closing — https://github.com/visionmedia/superagent/issues/1277#issuecomment-332242818

PR #1279, opened 2017-09-21
docs(piping): add note about chunked transfer encoding

jedwards1211:
I spent several hours today figuring out why I was unable to pipe a file in a request to a Flask (or any Python WSGI) server, so I figured I would share that knowledge with everyone else.


pornel:
Thanks

Issue #1288, opened 2017-10-02
Error: Attempt to change mapping for "form" extension from "application/x-www-form-urlencoded" to "application/x-www-form-urlencoded". 


sdesalas:
Superagent has [recently been upgraded](https://github.com/visionmedia/superagent/commit/96f3df844d6db1eb3d4024132a5d5ef73b13d801#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R34) to use version `2.0.3` of [mime](https://github.com/broofa/node-mime).

However, between version `1.4.1` and `2.0.3` of `mime` an [earlier warning](https://github.com/broofa/node-mime/blob/v1.4.1/mime.js#L26) has been upgraded [to an error](https://github.com/broofa/node-mime/blob/master/Mime.js#L32) when defining the same mime types twice .

And since many other packages and tools rely on `superagent`, the lack of pre-check before defining mime-types a second time is causing subsequent `require("superagent")` calls from erroring out as per the example below (whenever two packages feature `superagent` as a dependency).

Use following steps to reproduce the error.

```sh
npm install swagger-tools supertest
node -e 'require("swagger-tools"); require("supertest");'
```

And the output:

```
/home/user/projects/mytest/node_modules/mime/Mime.js:32
          throw new Error(`Attempt to change mapping for "${ext}" extension from "${this._types[ext]}" to "${type}". Pass \`force=true\` to allow this, otherwise remove "${ext}" from the list of extensions for "${type}".`);
          ^

Error: Attempt to change mapping for "form" extension from "application/x-www-form-urlencoded" to "application/x-www-form-urlencoded". Pass `force=true` to allow this, otherwise remove "form" from the list of extensions for "application/x-www-form-urlencoded".
    at Mime.define (/home/user/projects/mytest/node_modules/mime/Mime.js:32:17)
    at Object.<anonymous> (/home/user/projects/mytest/node_modules/supertest/node_modules/superagent/lib/node/index.js:71:6)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/user/projects/mytest/node_modules/supertest/lib/test.js:5:15)
```

sdesalas:
Fix as follows:

[...]

PR #1289, opened 2017-10-02
Fix for error in subsequent require() calls

sdesalas:
Superagent has recently been upgraded to use version 2.0.3 of mime.

However, between version 1.4.1 and 2.0.3 of mime an earlier warning has been upgraded to an error when defining the same mime types twice .

See https://github.com/visionmedia/superagent/issues/1288 for a longer explanation


pornel:
Thank you

sdesalas:
:+1:

sdesalas:
:man_dancing: ~~~~~~~

PR #1291, opened 2017-10-05
Catch and pass along errors in ok callback

### LINKED ISSUE #1247: Feature request: Customize error from .ok() callback
I have an API that responds to various failures with a HTTP 200 (long story), but responds with a field to key off of in the response body to determine the nature of the failure, something like:

``` json
{
  "stat": "fail",
  "code": 99,
  "message": "Something went wrong"
}
```

With superagent's `.ok()` callback, I can easily teach superagent what a failure response looks like:

``` js
.ok(res => res.ok && res.body.stat !== 'fail')
```

However, the error I get back from superagent in this case looks like `[Error: OK]` which is pretty confusing. I would like to be able to customize the error message returned to the caller, as well as maybe add additional helpful information such as the API's custom error code, etc.

I would totally be willing to provide a patch for this, but I first wanted to open this up to discussion on what the best implementation might be. Some ideas:

- Allow me to override the error message by setting `res.statusText` (or something) in the `.ok()` callback as soon as I decide to return falsy. [The client code would already allow this][1], but [the nodejs code does not][2]. node has [http.IncomingMessage#statusMessage][3] but then I have to set both properties in my `.ok()` callback (not quite ideal). Furthermore, this approach doesn't allow me to append any other properties to my error (like the above code, for example).
- Allow me to throw inside my `.ok()` callback, and if the surrounding code catches it use that error for `new_err` instead of creating one. This is a bit more of a significant change but it ultimately gives me complete control over the error if I want it.
- Provide a hook like `.ok()` that would allow me to customize how to create the error. I'm not sure this feature warrants a formal API, and given that the node and client implementations would have to be different anyways I'm not even sure this approach would work.

Thoughts?

[1]: https://github.com/visionmedia/superagent/blob/master/lib/client.js#L422
[2]: https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L682-L686
[3]: https://nodejs.org/api/http.html#http_message_statusmessage

jeremyruppel:
This patch wraps `_isResponseOk` in a try/catch block and passes along the caught error. The client version of superagent already appears to be doing this properly.

Fixes #1247

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -671,16 +671,20 @@ Request.prototype.callback = function(err, res){
   this.called = true;
 
   if (!err) {
-    if (this._isResponseOK(res)) {
-      return fn(err, res);
-    }
+    try {
+      if (this._isResponseOK(res)) {
+        return fn(err, res);
+      }
 
-    var msg = 'Unsuccessful HTTP response';
-    if (res) {
-      msg = http.STATUS_CODES[res.status] || msg;
+      var msg = 'Unsuccessful HTTP response';
+      if (res) {
+        msg = http.STATUS_CODES[res.status] || msg;
+      }
+      err = new Error(msg);
+      err.status = res ? res.status : undefined;
+    } catch (new_err) {
+      err = new_err;
     }
-    err = new Error(msg);
-    err.status = res ? res.status : undefined;
   }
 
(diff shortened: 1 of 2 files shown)


pornel:
That's very useful. Thank you.

PR #1296, opened 2017-10-12
Don't pop last array element when parsing a header.

### LINKED ISSUE #1295: Superagent "pops" valid header from response.
E.g. this code for me (chrome 61.0.3163.91)
```js
      const header = resp.xhr.getAllResponseHeaders();
      console.log(JSON.stringify(header));
      console.log(resp.header);
```
Prints this:
```
"content-type:\ttext/plain; charset=utf-8\nrange:\tbytes=0-524288"
{content-type: "text/plain; charset=utf-8"}
```

And this line of code suggests there should be a trailing `\r\n`:

https://github.com/visionmedia/superagent/blob/092e3d6d289f90af2551540a5fbcdf771ca9b7fe/lib/client.js#L223

Well, there is none as you can see in the snippets above. It just pops the actual header line...

nsf:
`xhr.getAllResponseHeaders()` may actually return a header string without trailing `\r\n`.

Fixes #1295.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -220,11 +220,12 @@ function parseHeader(str) {
   var field;
   var val;
 
-  lines.pop(); // trailing CRLF
-
   for (var i = 0, len = lines.length; i < len; ++i) {
     line = lines[i];
     index = line.indexOf(':');
+    if (index === -1) { // could be empty line, just skip it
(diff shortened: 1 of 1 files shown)


pornel:
Thank you!

PR #1298, opened 2017-10-17
Document that 3xx status codes are errors

mickeyreiss-visor:
According to the discussion in #283 as well as [the code in the latest release](https://github.com/visionmedia/superagent/blob/v3.7.0/lib/request-base.js#L230), 3xx responses are considered errors by default.

This PR increases clarity about this point in the docs.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -556,7 +556,7 @@ An "error" event is also emitted, with you can listen for:
 
       });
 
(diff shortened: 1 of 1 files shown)


pornel:
That is not quite accurate. 3xx by default are considered *redirects* and don't cause errors. They become errors only in non-default case where you disable/limit number of redirects.

mickeyreiss-visor:
Ah. In that case, what would be a good way to document the behavior for `304 Not modified`?

pornel:
Oh, I've forgotten about that one. Yes, it is an error indeed (we're unable to get its body). 

Perhaps clarify with "non-redirecting 3xx"

mickeyreiss-visor:
I pushed another version of the copy. Please take a look. ✍️ 👀

pornel:
That's good. Thank you.

PR #1301, opened 2017-10-19
ES6 in Node version

pornel:


pornel:
For browsers the major blocker is IE and a bit of iOS9 and Android 4. We have IE in CI, so hopefully this will catch any "leakage" of ES6 to the browser version.

PR #1308, opened 2017-11-07
bump node-mime to 2.x

killmenot:


pornel:

Unfortunately we're not ready to upgrade this yet. Mime 2 supports only Node 6+, and we still have users on Node 4.

I'm going to leave that pull request open until next major update of superagent when we'll be able to bump requirements to Node 6.

PR #1311, opened 2017-11-12
Do not parse json-seq as json

leedm777:
There's an odd mimetype `json-seq` which is a delimited series of JSON
objects (see [RFC 7464](https://tools.ietf.org/html/rfc7464)).

This format is not parseable JSON, so `isJSON('application/json-seq')`
should return `false`.


pornel:
Thank you

PR #1318, opened 2017-11-24
Trim response text before parsing as JSON.

jlek:
Hello,

One of our internal APIs adds whitespace before returning a JSON list. (i.e. `' ["foo", "bar"]'`). I've been able to make it work by copying your JSON parser and trimming the response text.

I thought that perhaps this tweak could benefit others as well, so I made this pull request. Hope it's okay. :)

Cheers,
John

### DIFF
--- a/lib/node/parsers/json.js
+++ b/lib/node/parsers/json.js
@@ -8,7 +8,7 @@ module.exports = function parseJSON(res, fn){
   });
   res.on('end', () => {
     try {
-      var body = res.text && JSON.parse(res.text);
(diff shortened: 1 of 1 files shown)


pornel:
Thank you for the PR, but I think it's not an appropriate modification. `JSON.parse` defines what JSON is, so if it doesn't parse, it's not a valid JSON.

superagent supports replacing parsers (globally via `supeagent.parse[mime]` or per request `.parse()`) so I'd recommend using that instead.

PR #1321, opened 2017-12-08
Fix documentation around attaching Buffer

### LINKED ISSUE #1198: Inconsistent documentation for .attach()
[This](https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L166) doesn't work in node:

```
.attach(new Buffer('<b>Hello world</b>'), 'hello.html')
```

As specified below, the first parameter of `.attach()` must be the field name. So this should be correct (at least it works for me):

```
.attach('someFieldName', new Buffer('<b>Hello world</b>'), 'hello.html')
```

Moreover, [`new Buffer(string)` is deprecated](https://nodejs.org/api/buffer.html#buffer_new_buffer_string_encoding) and should be replaced by `Buffer.from(string)`.

See also #762.

gabegorelick:
Closes #1198

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -164,7 +164,7 @@ RequestBase(Request.prototype);
  *
  * ``` js
  * request.post('http://localhost/upload')
- *   .attach(new Buffer('<b>Hello world</b>'), 'hello.html')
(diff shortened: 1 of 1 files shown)


pornel:
Thank you

PR #1326, opened 2017-12-19
Replace .catch with ['catch'] for IE9 Support

notflip:


notflip:
@pornel Is it possible to merge this? the ['catch'] is forward compatible! I tested it in non IE9 browsers. Thank you!

PR #1329, opened 2018-01-03
[node] Emit progress event while uploading Buffer via send method

sergey-akhalkov:
Hi!

Superagent is awesome and it allows to upload node Buffers like this `req.send(buffer)...` - this is exactly what we need. Unfortunately, `.on('progress', ...)` does not work in this case, this PR solves the problem.

Please let me know if you have any questions/considerations or see any issues.


kornelski:
Thank you!

Issue #950, opened 2016-03-21
Unbuffered parsers are tricky to use


kornelski:
Callback is fired in: https://github.com/visionmedia/superagent/blob/2910a198b7ca61c83eb308f67aba793d98d44ef8/lib/node/index.js#L803

which happens **before** async parsers set the body: https://github.com/visionmedia/superagent/blob/2910a198b7ca61c83eb308f67aba793d98d44ef8/lib/node/index.js#L787

so `res.body` in the callback ends up being the default `{}`. 

The parsers listen to `end` event. I suspect this may also cause race conditions with other code that listens to the `end` event and wants to use parsed body.

I think the right solution would be to wait with the callback until async parser is done (users who want to see response early can listen for the `response` event).

BTW: node code for choosing the right parser seems to be unnecessarily complicated (and uses both `parse` & `parser` variables).

pornel:
I think buffered responses should be the default, and unbuffered responses should be only opt-in. 

Having buffering mime-dependent is weird, especially that unknown types get unbuffered behavior. 

BTW, currently node and browser have different interfaces for parsers. There may be a possibility of unifying them, while sorting out buffering in the new parser API.

PR #1337, opened 2018-01-25
Allow default buffer settings based on response-type.

shreyjain1994:
I often use superagent to make requests to API's returning XML responses. In order to parse the response body, I need to call the buffer() method everytime, which gets tedious. 

I added a buffer field on the superagent object so that you can define certain response-types to be buffered or not buffered by default. I don't believe that it is a breaking change, and I think it might be useful to both me and other users of superagent.

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -852,7 +852,7 @@ Request.prototype._end = function() {
       unzip(req, res);
     }
 
-    if (buffer === undefined && exports.buffer[mime] !== undefined){
+    if (buffer === undefined && mime in exports.buffer){
       buffer = !!exports.buffer[mime];
     }
 
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -108,6 +108,15 @@ exports.serialize = {
 
 exports.parse = require('./parsers');
 
+
+/**
+ * Default buffering map. Can be used to set certain
+ * response types to buffer/not buffer.
+ *
+ *     superagent.buffer['application/xml'] = true;
+ */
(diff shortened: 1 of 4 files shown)


kornelski:
Technically it is a breaking change (#950). Your case didn't fail tests only because we don't use that MIME type in tests.

lib/node/index.js
```diff
@@ -843,6 +852,10 @@ Request.prototype._end = function() {
       unzip(req, res);
     }
 
+    if (buffer === undefined && exports.buffer[mime] !== undefined){
```
shreyjain1994: Are you sure it would be a breaking change? This block would only run if a user defined a mime type on `exports.buffer`, and since `exports.buffer` didn't exist until now, previously written code should run as normal. Also, this block still respects the fact that the `buffer()` method takes precedent over everything else.

kornelski: Ah, I see. In that case it's OK. 

However, have you considered https://visionmedia.github.io/superagent/#agents-for-global-state `agent.buffer(true)`?

akoptelov:
@kornelski Shoudn't buffering be turned on if a parser is defined? Otherwise I don't see any good in having a custom parser that has no means to set its result to the response...

shreyjain1994:
There should be no problems with `agent.buffer(true)` since it just calls the main `request.buffer` method internally.

However, I've added all necessary tests, including ones that show that `agent.buffer(true)` remains unaffected..

kornelski:
Thank you for adding the tests.

PR #1360, opened 2018-03-31
Add flags for 201 & 422 response statuses

itsfadnis:
Set flags for response statuses `201` & `422`, since they are also quite commonly used.


kornelski:
Thanks

PR #1371, opened 2018-04-30
Access cookie in headers correctly

liamkeaton:
Cookies laid in a previous request were not being merged with a new request due to the header being `Cookie` rather than `cookie`. This uses the internal `_headers` object where keys are coerced to be lowercase to fix this.


kornelski:
Good catch! Thanks.

Issue #1374, opened 2018-05-14
Uploads get aborted while in progress


DenisVuyka:
We experience a serious issue with the library (3.8.2) with regards to the Upload functionality. 

For large files or slow networks the requests get aborted on the timeout while upload technically is in progress. This can be easy reproduced when with Chrome DevTools and network throttling, on "Slow 3G" the 40 mb file gets aborted at the beginning (I believe after 6-10 seconds).

Uploading the same file using vanilla JS and network settings works fine. Upload can keep up for half an hour or so.

I believe the issue might be related to how you handle the "responseTimeoutTimer" and the fact it is not reset for the "upload.onprogress" (https://github.com/visionmedia/superagent/blob/master/lib/client.js#L692)

Issue #1244, opened 2017-07-03
progress bar shows 100% and it took minutes before uploading to server


vuthuytrangfr:
I tried to upload 326MB size file without chunking by superagent from android device (Xperia). The progress bar show 100% and after that it took 10 minutes or longer to upload file to server, send it to S3 and show the upload message.
I am search a solution for shorten the waiting time from showing 100% to showing the finished message.  Anyone know the reason and solution?

vuthuytrangfr:
video size: 326MB
- upload 100%: １６m：１８s
- after 14:31 s
error network
progressbar: back to 0%

pornel:
We don't measure progress. We only pass back what your browser reports, so misreporting of 100% is a browser/Android problem.

As for the error, check whether your server allows large files. Most servers and environments like PHP by default reject files larger than 20MB or so.

vuthuytrangfr:
thanks for the answer, I am using nginx server with ruby, react, and allow file larger than 150MB.
Do you know what is the problem of browser/Android or how to find out the reason?

vuthuytrangfr:
I have a problem something likes this 
https://github.com/rack/rack/issues/1075

titanism:
Browser upload progress is forwarded XHR loaded/total; 100% means the body left Android, not that nginx/Rack/S3 finished parsing, storing, or responding. Investigate server processing/limits.

v10.4.0 is available on npm: https://www.npmjs.com/package/superagent/v/10.4.0
GitHub release: https://github.com/forwardemail/superagent/releases/tag/v10.4.0

Consider supporting our efforts to maintain this project by using Forward Email at https://forwardemail.net.  We run the only 100% open source and 3rd party audited email service in the world.

titanism:
Closing this for now. If this still affects v10.4.0, please open a new issue with a current reproduction.

v10.4.0 is available on npm: https://www.npmjs.com/package/superagent/v/10.4.0
GitHub release: https://github.com/forwardemail/superagent/releases/tag/v10.4.0

Consider supporting our efforts to maintain this project by using Forward Email at https://forwardemail.net.  We run the only 100% open source and 3rd party audited email service in the world.

Issue #1115, opened 2016-11-23
Powerful timeouts


kornelski:
Currently there's a single `timeout` option that sets a deadline for the entire request to complete. 

This is suboptimal for large files and slow connections, because it doesn't distinguish between the server not responding at all, and the file downloading reliably, but slowly.

In reality there are multiple stages of the request that can time out:

* DNS timeout
* TCP/IP connection timeout
* File upload
* Waiting for HTTP response headers
* Waiting for each chunk of HTTP response body

I'm thinking of hooking into more of these steps to allow having shorter timeouts as appropriate, e.g. DNS and TCP/IP connection never take more than a few seconds, even if the whole request needs minutes to download.

What should be the API for this? Should we try to interpret existing `.timeout(2000)` API in some clever way?

focusaurus:
👎 for clever API interpretations and overloading. My first thought would be leave `.timeout(2000)` as an alias for the entire request completing including the full body, and introduce APIs like `.resolveTimeout(500)`, `.connectTimeout(1000)`, `.headerTimeout(3000)`, `.chunkTimeout(4000)`.

pornel:
By overloading, do you mean `.timeout(Number)` vs `.timeout({connect:…, chunk:…})`?

pornel:
Also, should we have any defaults?

focusaurus:
Yes, overloading meaning a single API function call that has multiple meanings or behaviors depending on arguments provided. I think in general that's an antipattern and having distinct API functions that do 1 clear thing is preferable, even if it means the API surface is bigger (it's just being honest about the actual API surface).

I don't mind `.timeout({connect: 42, chunk: 97})` that much though. It's not terrible. It just requires good validation and I think it's easier to document and more obvious when you typo `.conectTimeout(42)` and get an obvious exception vs calling `.timeout({conect: 42})` and maybe having to debug why the timeout seems to not be taking effect as expected. If the code detects unexpected option names and throws an exception immediately, it's close enough I think.

I think we should have no defaults and do what the runtimes/node/oses do by default.

kornelski:
The MVP of response timeout seems to work well enough, so I'll leave it at that.

Issue #840, opened 2016-01-08
progress event not working for send file data event


jitchavan:
i am trying to use attach event but it is not working chrome. so i am using below code for file upload which is working fine.

But it does not supporting progress event on send event for file.  

```
    const req = request.post(`https://api.cloudinary.com/v1_1/${settings.CLOUDINARY_ACCOUNT_NAME}/image/upload`);

    const data = new FormData();
    files.forEach((file) => {
      data.append('file', file);
    });
    data.append('api_key', settings.CLOUDINARY_API_KEY);
    data.append('timestamp', Date.now() / 1000);
    data.append('upload_preset', settings.CLOUDINARY_UPLOAD_PRESET);

    req.send(data);
    req.on('progress', function(e) {
                $('#'+this.filename).width(e.percent+'%');
            });
    req.end((err, res) => {
      if (err) {
        console.log('error', err);
      } else {
        console.log('success', res);
      }
    });
```

pornel:
I think it's been fixed in 2.3.0

halt-hammerzeit:
I confirm that this has been fixed and is working for file upload in Chrome

PR #1375, opened 2018-05-14
keep upload alive for large files or slow connections

DenisVuyka:
Addresses issue #1374

Keeps upload alive for very large files or slow connections, resets deadline/timeout timers while data is still being sent over the wire.

I have tested the fix with network throttling using Chrome DevTools, also tested on a real project. However I cannot find a way to create a unit test that has slow upload with progress updates simulated.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -709,6 +709,14 @@ Request.prototype._end = function() {
 
   // progress
   const handleProgress = (direction, e) => {
+    // reset 'deadline' timer if still sending data
+    if (self._timer) {
+      clearTimeout(self._timer);
+    }
+    // reset 'timeout' timer if still sending data
+    if (self._responseTimeoutTimer) {
+      clearTimeout(self._responseTimeoutTimer);
+    }
     if (e.total > 0) {
       e.percent = e.loaded / e.total * 100;
     }


kornelski:
Thank you for the PR. However, I don't think that's appropriate. 

I see what you want — monitoring whether connection is still alive and timing out only after data stops flowing. That would be useful. But that's not what current timers are supposed to do. The deadline's main purpose *is* to kill slow connections, even if they're still going. 

If you can accept the upload taking half an hour, then you can't set the deadline (or you have to set it to half an hour).

The `_responseTimeoutTimer` is cleared in readyState change >= 2, which is fired when headers are received. Its not set again, so doesn't need to be cleared later.

So if you'd like to implement watchdog timer for connection stalling, then you need to add a third timer. Don't cancel the existing ones.

DenisVuyka:
I see what you mean, thanks. Will try to increase timeout settings at the client side.

eromano:
@kornelski also if we add a third timer the deadline timeout will be still active . 
Should I just disable the deadline timeout and other this third timeout?

kornelski:
No, don't touch the other options. They are independent.

kornelski:
From what I understand, you are setting the deadline timer, which sole purpose is to unconditionally abort the request, and then you don't like that the timer fires and unconditionally aborts the request.

There's nothing wrong with the deadline timer, it does exactly what it's supposed to do. It's just not a feature for you, so *don't use it*. Don't call `.timeout({deadline:…})`. 

> How can I avoid the deadline timeout to happen if I don't clear it?

You can't avoid the timer if it's already been set. By design supposed to be unconditional. If you change the implementation to clear the timer in some circumstances, then you just break that feature.

What I'm suggesting is that you add a third kind of timer, completely separate from the other two, and then:

* Do not set the deadline, at all. Never. Just forget it ever existed.
* Use the third kind of timer.

In that situation you don't need to clear the deadline timer.

eromano:
Ok I got it, thanks.
This timer will works only if the deadline timer is not set or it's value is minor of the deadline.
Do you would like to have this functionality in superagent or is better that we just overwrite it in our side?

kornelski:
Yes, this functionality (timeout since last bit of data was sent or received) would be useful in superagent. Here's previous discussion of that feature: #1115

I've just remembered we've had reports that the upload events are buggy in some browsers: #1244 #840 

So this is going to be a serious problem for uploads, and will require special handling (e.g. find a way to detect when the browser fires events for buffering rather than actual sending, or support the timeout only in browsers tested and verified to work).

eromano:
Ok thanks , I guess I will try to create a PR but I will sure need your help

kornelski:
Sure! I'm happy to help.

PR #1383, opened 2018-05-30
Update dependency 'cookiejar' to version 2.1.2

danielmcq:
This fixes an issue seen sometimes when using superagent in node.js. The an error is thrown when an agent is used and the server send invalid `Set-Cookie` headers. The package [x-ray](https://github.com/matthewmueller/x-ray), which relies on [x-ray-crawler](https://github.com/lapwinglabs/x-ray-crawler), which relies on superagent, has an example of this problem seen here: https://github.com/matthewmueller/x-ray/issues/234.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "component-emitter": "^1.2.0",
-    "cookiejar": "^2.1.0",
+    "cookiejar": "^2.1.2",
     "debug": "^3.1.0",
     "extend": "^3.0.0",
     "form-data": "^2.3.2",


kornelski:
Just FYI this isn't necessary. You can run `npm update` to achieve the same effect.

danielmcq:
Thank you for approving and merging. I just wanted to make sure the dependency version was explicit because any projects using a *package-lock.json* or that use *yarn.lock* won't automatically get updated dependencies. This way with the next version of superagent any other project that updates to use that version will be certain that the fix done in the cookiejar project will be included.

Issue #980, opened 2016-04-29
HTTP/2 support


warrenseine:
AFAIK, `superagent` relies on the standard Node module HTTP/HTTPS, which, AFAIK again, is not HTTP/2-compatible. Therefore, performing HTTP/2 requests with `superagent` is not possible for now.

Technically, the [`node-http2`](https://github.com/molnarg/node-http2) module could be used to support HTTP/2 without rewriting the whole thing. Is this part of a roadmap? Could there be an alternative I didn't think of?

pornel:
It hasn't been on my radar.

With `request-base` we've done some work to make code more reusable between browser and node, so this approach could be used to write a node-http2-based version as well.

Alternatively, if you had a node module that has the same interface as the built in `require('https')` module, you could swap these.

harshu108:
```
describe('Ping related', function() {
              
//make connect related variables as false.
               before(function(){
                     mailchimpDS.connected = false;
                      mailchimpDS.connecting = false;
                    
//Stub connect method here
                     connect_stub = sinon.stub(mailchimpDS.connector,'connect',function(callback){
                            
                            var someFunc = function(){
                                   callback(null);
                            }
                            process.nextTick(someFunc);
                     })
                  });
//restore connect stub
                  after(function(){
                     connect_stub.restore();
                 });

                  
         it('should successfully test connection', function(done) {
//stub send request
              create_stub_ping = sinon.stub(mailchimpDS.connector, 'sendRequest', function(req, callback) {
                  var output = {
                      "statusCode": 200,
                      "account_id": "8ba087e5f225fdd80ade5e36c",
                      "account_name": "persistent",
                      "email": "sonia_targe@persistent.co.in",
                      "role": "owner"
                  };
                  return callback(null, output);
              });
              
              
              mailchimpDS.connector.ping(function(err, body) {        
                  if (err) {
                      expect(err).to.not.be.null;
                  } else {
                      expect(body).to.be.undefined;
                  }
                  create_stub_ping.restore();
                 
                  done();
              });  
//Call connect
              mailchimpDS.connect(function(err, res) {});
            
          });
)};
```

warrenseine:
Thanks for the feedback. I'll try that.

In case this worked by swapping the modules, would a PR adding a [protocol](https://github.com/visionmedia/superagent/blob/master/lib/node/index.js#L71) and a new dependency would be considered?

pornel:
A new dependency for node is OK.

[...]

PR #1399, opened 2018-07-31
Support http2.

sogaani:
Hi.

This PR enable http/2, if EXPOSE_HTTP2 env is true.

This PR still not support ALPN(as @kornelski mentioned in https://github.com/visionmedia/superagent/issues/980).
I'd prefer to add ALPN functionality after merging this PR and [express http/2 PR](https://github.com/expressjs/express/pull/3390).


kornelski:
Thank you! That's a very useful functionality.

kornelski:
Does `EXPOSE_HTTP2` affect ability to make HTTP/1 requests? Is Node going to fall back automatically? I wonder if we need to give more runtime control over this.

lib/node/http2wrapper.js
```diff
@@ -0,0 +1,186 @@
...
+  const sessionOptions = Object.assign({}, options);
+  if (options.socketPath) {
+    sessionOptions.socketPath = options.socketPath;
+    sessionOptions.createConnection = this.createUnixConnection.bind(this);
+  }
+
+  this._headers = {};
+
+  const session = http2.connect(`${protocol}://${host}:${port}`, sessionOptions);
+  this.setHeader('host', `${host}:${port}`)
```
kornelski: Is it necessary here? In normal implementation we allow `options.host` to be different than host header, because it's useful to do `request.get('http://[ip address]').host('example.com')` to force request to connect to a particular IP address.

lib/node/http2wrapper.js
```diff
@@ -0,0 +1,186 @@
...
+ * Inherit from `Stream` (which inherits from `EventEmitter`).
+ */
+util.inherits(Request, Stream);
+
+Request.prototype.createUnixConnection = function (authority, options) {
+  switch (this.protocol) {
+    case 'http':
+      return net.connect(options.socketPath);
+    case 'https':
+      options.ALPNProtocols = ['h2'];
```
kornelski: AFAIK browsers also send http/1 here. Do we support a fallback?

lib/node/http2wrapper.js
```diff
@@ -0,0 +1,186 @@
...
+  const keys = Object.keys(headers);
+  const http2Headers = {};
+  for (var i = 0; i < keys.length; i++) {
+    let key = keys[i];
+    let value = headers[key];
+    key = key.toLowerCase();
+    switch (key) {
+      case HTTP2_HEADER_HOST:
+        key = HTTP2_HEADER_AUTHORITY;
+        value = value.startsWith('http') ? parse(value).host : value;
```
kornelski: Why is such conditional parsing needed?

test/node/pipe.js
```diff
@@ -12,13 +17,19 @@ app.get("/", (req, res) => {
 });
 
 app.post("/", (req, res) => {
-  res.send(req.body);
+  if (process.env.EXPOSE_HTTP2){
+    res.set('content-type', 'application/json');
```
kornelski: Why is this needed?

test/node/unix-sockets.js
```diff
@@ -94,6 +105,7 @@ describe("[unix-sockets] https", () => {
         .get(`${base}/request/path`)
         .ca(cert)
         .end((err, res) => {
+          console.log(err);
```
kornelski: 👀

test/node/user-agent.js
```diff
@@ -6,7 +6,7 @@ const base = setup.uri;
 
 describe("req.get()", () => {
   it("should set a default user-agent", () =>
-    request.get(`${base}/ua`).then(res => {
+    request.put(`${base}/ua`).then(res => {
```
kornelski: Can you explain this change?

test/support/server.js
```diff
@@ -17,7 +42,15 @@ app.all('/url', (req, res) => {
 });
 
 app.all('/echo', (req, res) => {
-  res.writeHead(200, req.headers);
+  let headers = req.headers;
+  if(process.env.EXPOSE_HTTP2){
+    Object.keys(headers).forEach(function(name) {
+      if(isPseudoHeader(name)){
+        delete headers[name];
```
kornelski: Maybe this should translate the header to HTTP/1 header, or allow both in the test using it?

lib/node/http2wrapper.js
sogaani: Current implement does not support fall back. I can implement fallback, I plan to take fallback after this PR. Or Should I integrate  with this PR?

kornelski:
@focusaurus do you have a preferences how HTTP/1 vs 2 API should look like?

lib/node/http2wrapper.js
sogaani: This is add for compatibility with Http/1. HTTP2_HEADER_AUTHORITY(alternative Host in http/2) only support domain string, otherwise it raise error. However, I met Host header with schema(like Host: http://example.com) in some express tests.

test/node/pipe.js
sogaani: http/2 module have not full compatibility with http/1 module. http/2 imcommingHttpRequest does not have body property.

test/node/unix-sockets.js
sogaani: My misstake! I will fix 👍

test/node/user-agent.js
sogaani: http/2 client module cannot send body with head or get Method. But, it seems to not need here. I check this again later.

test/support/server.js
sogaani: HTTP/2 use some headers to control protocol(e.g. :method, :authority, :path ...) and it can not set by user. I just delete those headers here.

lib/node/http2wrapper.js
sogaani: I think current implementation supports `request.get('http://[ip address]').host('example.com')`. I will check later 👍 .

focusaurus:
Hmm. Am I interpreting the discussion correctly? So if a user sets `EXPOSE_HTTP2=true` they can only connect to http2 servers from that entire node process and connections to http/1 servers will fail? That sounds untenable. I think I'd prefer leaving HTTP/1 as the default and exposing an API to use http/2 on a per-request basis. For that I think the `.http2()` method name is a good API.

lib/node/http2wrapper.js
kornelski: So it would be better to improve the check, otherwise host like `httparchive.org` would be misparsed :)

```js
/^https?:\/\//.test(value)
```

sogaani:
@focusaurus Thank you for feedback!.
>  I think I'd prefer leaving HTTP/1 as the default and exposing an API to use http/2 on a per-request basis. For that I think the .http2() method name is a good API.

It is reasonable. I change this soon. 👍

test/node/user-agent.js
sogaani: I test locally and noticed that this change is not needed. I will revert this change.

lib/node/http2wrapper.js
sogaani: Still not send header here and we can change `Host` header by `set` method. I add test for this.

lib/node/http2wrapper.js
kornelski: I see. If there's no fallback, then that's OK.

sogaani:
I addressed review and change the API to exposing an HTTP/2 API.

Superagent can be enabled http2 by http2 field:

```js
    const request = require('superagent');
    request.http2 =true;
    request
      .get('http://example.com/search')
      .then(res => {

      });
```

A request can be enabled http2 by http2 method:

```js
    const request = require('superagent');
    request
      .get('http://example.com/search')
      .http2()
      .then(res => {

      });
```

sogaani:
ping @kornelski  @focusaurus

kornelski:
Thanks

focusaurus:
Awesome contribution @sogaani !

PR #1413, opened 2018-09-13
fix req.serialize in node

### LINKED ISSUE #1141: .serialize() is not working in node agent
```js
superagent.
  post('/').
  set('Content-Encoding', 'gzip').
  serialize(function (d) {return require('zlib').gzipSync(JSON.stringify(d))}).
  send({send: {me: 'gzipped'}}).
  end(function () {})
```
this sends invalid gzip (probably JSON stringified)

snoguchi:
`request.serialize(fn)` is not working in Node.  This patch uses user specified `_serializer` in case it exists.

Fixes #1141


kornelski:
Thanks!

PR #1426, opened 2018-10-20
Added missing semicolons

reviewher:
More generally, there seems to be no standard lint tool (ESLint, JSHint). They would have been able to pick up issues like these.


kornelski:
Thanks for the PR.

I agree it's better to have some consistent style, but such changes for sake of changes are just going to cause merge conflicts with other PRs.

PR #1438, opened 2018-11-25
Adding link to new cheerio plugin

mmmmmrob:
Simple plugin that adds cheerio to html responses, useful when testing server-side rendering with supertest.


kornelski:
Thanks

PR #1449, opened 2018-12-24
There is a bug in JSON parsing

DaggieBlanqx:
I am using a WordPress API module in NodeJS ( https://github.com/WP-API/node-wpapi )
When I make API requests ; it gets the json data , but causes an error .
When I open the stacktrace :

{ SyntaxError: Unexpected token ﻿ in JSON at position 0 at JSON.parse (<anonymous>) at IncomingMessage.res.on (C:\cde\fn\notification-stream\node_modules\superagent\lib\node\parsers\json.js:11:35) at IncomingMessage.emit (events.js:187:15) at endReadableNT (_stream_readable.js:1092:12) at process._tickCallback (internal/process/next_tick.js:63:19)

When I examine the file superagent/lib/node/parsers/json.js , I found a bug in how it parses JSON .

I added the .trim() to ensure that there is no space in the string just before it is parsed

### DIFF
--- a/lib/node/parsers/json.js
+++ b/lib/node/parsers/json.js
@@ -8,7 +8,7 @@ module.exports = function parseJSON(res, fn){
   });
   res.on('end', () => {
     try {
-      var body = res.text && JSON.parse(res.text);
+      var body = res.text.trim() && JSON.parse(res.text.trim());
     } catch (e) {
       var err = e;
       // issue #675: return the raw response if the response parsing fails


DaggieBlanqx:
I noticed the bug when working with WordPress REST API and running this on a UNIX computer .

It seems like a minor issue but trust me my wp-json data is good , I linted it during the testing but still realised that the issue was with your module not trimming strings when needed.

DaggieBlanqx:
Again your module assumes that data coming in is compliant to RFC standards , so you ignored the importance of  trimming it .

Trim a string before parsing it 😂😂😂 else you're setting up the code for vulnerabilities .

kornelski:
No, that's not how this works. You must fix your WordPress to send valid data, rather than break everything touching the broken WordPress.

PR #1456, opened 2019-01-31
Update json.js

dylanzhang123456:
when response is a text message, this code will cause defect,error message will be like:
{ SyntaxError: Unexpected token h in JSON at position 0
...
..
}

please help review,Thanks.


niftylettuce:
Could you please update your pull request based of latest `master` branch?

Issue #1466, opened 2019-03-11
Regression in v4: expect(404) treats status code 404 as an error


bajtos:
Consider the following code snippet:

```js
const supertest = require('supertest');

supertest('http://example.com')
  .get('/path-not-found')
  .expect(404)
  .then(
    ok => console.log('PASS'),
    err => {
      console.error('FAIL', err.stack);
      process.exit(1);
    });
```

Run it using `supertest@3.4.2` to get the expected outcome:

```
PASS
```

Run it using `supertest@4.0.0` to reproduce the bug:

```
FAIL Error: Not Found
    at Test.Request.callback (/private/tmp/bug/node_modules/supertest/node_modules/superagent/lib/node/index.js:804:15)
    at IncomingMessage.parser (/private/tmp/bug/node_modules/supertest/node_modules/superagent/lib/node/index.js:1036:18)
    at Stream.emit (events.js:182:13)
    at Unzip.unzip.on (/private/tmp/bug/node_modules/supertest/node_modules/superagent/lib/node/unzip.js:55:12)
    at Unzip.emit (events.js:187:15)
    at endReadableNT (_stream_readable.js:1094:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

bajtos:
I have reviewed the changes in supertest and superagent, but not was able to find the source of the problem.

- supertest changes from v3 to v4: https://github.com/visionmedia/supertest/compare/v3.4.2...v4.0.0
- superagent changes from v3 to v4: https://github.com/visionmedia/superagent/compare/v3.8.3...v4.1.0

dejanr:
I noticed the same this morning, it must be something related to promises, as by using the callback it works.

raymondfeng:
I can confirm the issue is introduced by superagent instead of supertest.

raymondfeng:
The root cause seems to be https://github.com/visionmedia/superagent/compare/v3.8.3...v4.0.0#diff-be27821c599006faab0d29e2cace0392R238. Commenting it out fixes our issue.

midrissi:
Actually, this is the commit that causes the issue: https://github.com/visionmedia/superagent/commit/6670a0fb708fef3d8e09538ee8af6c73f682f346#diff-be27821c599006faab0d29e2cace0392

holm:
This was definitely surprising to us. It seems to be listed as a "minor change" int he history, but its not easy to understand what "all error events" mean.

[...]

PR #1468, opened 2019-03-12
Fix handling of FormData errors

bajtos:
This patch is fixing regression introduced by https://github.com/visionmedia/superagent/commit/6670a0fb708fef3d8e09538ee8af6c73f682f346, see https://github.com/visionmedia/superagent/issues/1466 for detailed discussion.

/cc @kornelski  @lapo-luchini

I have verified the fix by running the following test manually. ~~Please advise me how to capture these test cases in your automated test suite.~~

The first two use cases are already captured in the existing test suite, so I updated the existing tests to match the new behavior. Then I added two new tests to verify what happens when two errors are triggered by a single request (the last case below).

```js
// remote host is listening, attaching a missing file
// PROMISE VARIANT
require('.').post('http://example.com/does-not-exist')
.field({a:1,b:2})
.attach('c', 'does-not-exists.txt')
//.on('error', e => console.log('ERR', e.message))
.then(ok => console.log('THEN'))
.catch(e => console.log('CATCH', e.message));

// remote host is listening, attaching a missing file
// CALLBACK VARIANT
require('.').post('http://example.com/does-not-exist')
  .field({a:1,b:2})
  .attach('c', 'does-not-exists.txt')
  //.on('error', e => console.log('ERR', e.message))
  .end((err, res) => {
    if (err) console.log('CATCH', err && err.message);
    else console.log('THEN');
  });

// remote host is not listening
// two errors are triggered: cannot read the file, cannot connect to 127.0.0.1:80
require('.').post('http://127.0.0.1/test')
.field({a:1,b:2})
.attach('c', 'does-not-exists.txt')
//.on('error', e => console.log('ERR', e.message))
.then(ok => console.log('THEN'))
.catch(e => console.log('CATCH', e.message));
```

### DIFF
--- a/lib/node/index.js
+++ b/lib/node/index.js
@@ -264,7 +264,13 @@ Request.prototype._getFormData = function() {
   if (!this._formData) {
     this._formData = new FormData();
     this._formData.on('error', err => {
-      this.emit('error', err);
+      debug('FormData error', err);
+      if (this.called) {
+        // The request has already finished and the callback was called.
+        // Silently ignore the error.
+        return;
+      }
+      this.callback(err);
       this.abort();
     });
   }
--- a/lib/request-base.js
+++ b/lib/request-base.js
@@ -242,7 +242,6 @@ RequestBase.prototype.then = function then(resolve, reject) {
       console.warn("Warning: superagent request was sent twice, because both .end() and .then() were called. Never call .end() if you use promises");
     }
     this._fullfilledPromise = new Promise((innerResolve, innerReject) => {
-      self.on('error', innerReject);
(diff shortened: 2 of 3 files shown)


bajtos:
I have added two new tests to verify that ECONNRESET errors take precedence over ENOENT errors, I think we have sufficient test coverage now. I also fixed `npm test` failures and improved the commit message to better describe what is being changed.

As far as I am concerned, this pull request is ready for final review and merging.

@kornelski Could you please take a look? I understand you might not have time to do it right now, in which case could you please let me know when you will have time to process my pull request?

PR #1471, opened 2019-03-23
Invalid redirect in pipe

louisbuchbinder:
There is a bug in redirects during a pipe. If the request responds with a redirect status, but an empty Location header then superagent will throw `Uncaught TypeError: Cannot read property '_pipeContinue' of undefined`.

To reproduce you can checkout my tests (without the patch) and run them. You will see:
```bash
 330 passing (8s)                               
  5 pending                                
  1 failing                                      
                                             
  1) request pipe should not throw on bad redirects:          
     Uncaught TypeError: Cannot read property '_pipeContinue' of undefined
      at ClientRequest.req.once.res (lib/node/index.js:414:33)
      at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:543:21)
      at HTTPParser.parserOnHeadersComplete (_http_common.js:112:17)
      at Socket.socketOnData (_http_client.js:440:20)
      at addChunk (_stream_readable.js:263:12)
      at readableAddChunk (_stream_readable.js:250:11)
      at Socket.Readable.push (_stream_readable.js:208:10)
      at TCP.onread (net.js:597:20)  
```

or you can reproduce with by hand with something like:

```
http.createServer((req, res) => {
  res.setHeader('Location', '');
  res.statusCode = 302;
  res.end();
}).listen(8000);

require('superagent').get('localhost:8000/').pipe(require('fs').createWriteStream('./temp'));

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ coverage
 .nyc_output
 lib
 dist
+*.swp
--- a/test/node/pipe.js
+++ b/test/node/pipe.js
@@ -17,8 +17,16 @@ app.get('/', (req, res) => {
   fs.createReadStream('test/node/fixtures/user.json').pipe(res);
 });
 
-app.post('/', (req, res) => {
-  if (process.env.HTTP2_TEST) {
+app.get("/redirect", (req, res) => {
+  res.set('Location', '/').sendStatus(302);
+});
+
+app.get("/badRedirectNoLocation", (req, res) => {
+  res.set('Location', '').sendStatus(302);
+});
+
+app.post("/", (req, res) => {
+  if (process.env.HTTP2_TEST){
     // body-parser does not support http2 yet.
     // This section can be remove after body-parser supporting http2.
     res.set('content-type', 'application/json');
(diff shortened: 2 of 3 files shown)


test/node/pipe.js
```diff
@@ -92,4 +100,48 @@ describe("request pipe", () => {
     });
     req.pipe(stream);
   });
+
+  it("should follow redirects", done => {
```
louisbuchbinder: added a test for pipes with a good redirect

test/node/pipe.js
```diff
@@ -92,4 +100,48 @@ describe("request pipe", () => {
...
+      JSON.parse(fs.readFileSync(destPath, "utf8")).should.eql({
+        name: "tobi",
+      });
+      responseCalled.should.be.true();
+      done();
+    });
+    req.pipe(stream);
+  });
+
+  it("should not throw on bad redirects", done => {
```
louisbuchbinder: added a test to prove that redirects with a missing Location will not throw and error.

.gitignore
```diff
@@ -9,4 +9,5 @@ components
 test/node/fixtures/tmp.json
 .idea
 superagent.js
-package-lock.json
\ No newline at end of file
+package-lock.json
+*.swp
```
louisbuchbinder: added `*.swp` to the gitignore for vim files

louisbuchbinder:
failing tests in node10 and 11... updating

test/node/pipe.js
louisbuchbinder: I realize now there is already a test for a good redirect in `./test/node/pipe-redirect.js`

lib/node/index.js
```diff
@@ -409,14 +409,12 @@ Request.prototype.pipe = function(stream, options){
 Request.prototype._pipeContinue = function(stream, options){
   this.req.once('response', res => {
     // redirect
-    const redirect = isRedirect(res.statusCode);
-    if (redirect && this._redirects++ != this._maxRedirects) {
-      if (!res.headers.location) {
-        this.callback(new Error('No location header for redirect'), res);
-        stream.end();
```
louisbuchbinder: On second thought, superagent should not be responsible for ending the client stream if there is an error. Instead the client should listen for the `error` event and then handle their stream on their own.

louisbuchbinder: please advise...

louisbuchbinder:
@niftylettuce I rebased and update for the linting rules

niftylettuce:
@louisbuchbinder v5.0.4 is released, thank you

PR #1493, opened 2019-06-07
fix: unify "undefined" password treatment for auth()

jalooc:
Using `.auth('tobi')` gives different results than `.auth('tobi', undefined)`. This happens, because in the first case `pass` is [replaced with an empty string based on the args count](https://github.com/visionmedia/superagent/blob/master/src/node/index.js#L574), whilst not being replaced in the second case, [gets stringified](https://github.com/visionmedia/superagent/blob/master/src/request-base.js#L463) to the `'undefined'` string.

Not sure if this is an intended behavior, but imho it's misleading, because in both cases the `pass` is `undefined`. The only difference, is that in the first case it's implicit and explicit in the second one.

This PR unifies treatment of the `undefined` `pass`.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -553,7 +553,7 @@ Request.prototype.accept = function(type) {
  */
 
 Request.prototype.auth = function(user, pass, options) {
-  if (arguments.length === 1) pass = '';
+  if (pass === undefined) pass = '';
   if (typeof pass === 'object' && pass !== null) {
     // pass is optional and can be replaced with options
     options = pass;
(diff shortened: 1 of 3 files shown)


niftylettuce:
@jalooc 	let me know!

jalooc:
Yes! Sorry, I missed the message. I'll handle it this week.

niftylettuce:
@jalooc 	would love this ❤️

niftylettuce:
@jalooc 	any interest in resolving this? 👍 😄

niftylettuce:

PR #1495, opened 2019-06-13
Disable setting default user-agent

### LINKED ISSUE #1479: how to avoid sending user-agent header?
In [unirest](https://github.com/Kong/unirest-nodejs), when I don't specify `user-agent` header, it's just not sent. However `superagent` sends `node-superagent/5.0.2`.

`.set('user-agent', undefined)` fails with `TypeError [ERR_HTTP_INVALID_HEADER_VALUE]: Invalid value "undefined" for header "user-agent"`

`.set('user-agent', '')` sends `{ 'user-agent': '' }`

`.set('user-agent', null)` sends `{ 'user-agent': 'null' }`

apoorv-mishra:
Fixes #1479

### DIFF
--- a/src/node/index.js
+++ b/src/node/index.js
@@ -21,7 +21,6 @@ const semver = require('semver');
 const safeStringify = require('fast-safe-stringify');
 
 const utils = require('../utils');
-const pkg = require('../../package.json');
 const RequestBase = require('../request-base');
 const { unzip } = require('./unzip');
 const Response = require('./response');
@@ -132,14 +131,11 @@ exports.buffer = {};
  * @api private
  */
 function _initHeaders(req) {
-  const ua = `node-superagent/${pkg.version}`;
   req._header = {
     // coerces header names to lowercase
-    'user-agent': ua
   };
   req.header = {
     // preserves header name case
-    'User-Agent': ua
   };
 }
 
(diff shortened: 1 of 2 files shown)


niftylettuce:
Can you please fix the tests @apoorv-mishra?

apoorv-mishra:
> Can you please fix the tests @apoorv-mishra?

@niftylettuce I have a bit of difficulty here. I assumed that we would need to remove tests which test for setting default `user-agent`. Tests for that are in `test/node/user-agent.js`. I removed it and ran `npm run build && npm run lint && make test`. Still tests seem to fail. What should be done here?

apoorv-mishra:
@niftylettuce Updated.

niftylettuce:
I've released this as v5.1.0 to npm, it has a minor version bump because existing users may have issues with the user-agent being stripped in this version.

`npm install superagent@latest`

https://github.com/visionmedia/superagent/releases/tag/v5.1.0

wyardley:
IMHO, this would have been better as a breaking / major change.
Certain sites (GitHub, for example) will reject requests sent without a `User-Agent` header.

comster:
> IMHO, this would have been better as a breaking / major change.
> Certain sites (GitHub, for example) will reject requests sent without a `User-Agent` header.

For instance this does break https://github.com/lambtron/get-stargazers-emails - FWIW

Issue #1498, opened 2019-06-20
Please update HISTORY for version 5!


mikemaccana:
https://github.com/visionmedia/superagent/blob/master/HISTORY.md goes up to 4.x and stops. Would love to see what's changed!

niftylettuce:
I am now using Releases tab of GitHub

https://github.com/visionmedia/superagent/releases

niftylettuce:
I was hoping to get admin by @tj on this repo so I can fix the GH-pages

mikemaccana:
Fair enough. Can you merge the PR above?

niftylettuce:
done!! thank you again!!!!!!

PR #1499, opened 2019-06-21
Add reference to GitHub releases page

mikemaccana:
Per https://github.com/visionmedia/superagent/issues/1498#issuecomment-504126056


niftylettuce:
@mikemaccana thank you so much for doing this.  this is incredibly helpful, it's hard to keep up at times!

PR #1500, opened 2019-07-14
Update README.md

abhisheq6u:
I just noticed, someone probably misspelled "supporting" as "sporting" in README.md


niftylettuce:
Both are interchangeable here, but supporting is probably better suited.  Thanks for the PR.

PR #1515, opened 2019-10-03
feat: add secure cookie override to agent

ejose19:
Allow overriding the default secure cookie for Cookie access info (use case: doing tests directly with express / koa app without nginx / tls certs)

### DIFF
--- a/src/agent-base.js
+++ b/src/agent-base.js
@@ -24,7 +24,8 @@ function Agent() {
   'key',
   'pfx',
   'cert',
-  'disableTLSCerts'
+  'disableTLSCerts',
+  'sendSecureCookie'
 ].forEach(fn => {
   // Default setting for all requests from this agent
   Agent.prototype[fn] = function(...args) {
--- a/src/node/agent.js
+++ b/src/node/agent.js
@@ -50,6 +50,10 @@ function Agent(options) {
     if (options.rejectUnauthorized === false) {
       this.disableTLSCerts();
     }
+
+    if (options.sendSecureCookie) {
+      this.sendSecureCookie();
+    }
   }
 }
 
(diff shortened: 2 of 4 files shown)


ejose19:
Worth noting, I had to bump the engine -> node to 7.0.0 because I was getting a version error from http2 wrapper.

Also, this commit goes in hand with this [commit](https://github.com/ejose19/supertest/commit/2e1e35d8718f21028ba31bbfdade71b4b2f5ffe9) for supertest, on which I will wait for this PR to be merged first as the tests for supertest commit requires proposed  changes.

End usage:
```
const agent = supertest.agent(appCB, { secureCookie: true });
```

niftylettuce:
Can you please add a test?  Also it looks like there is a merge conflict that will need fixed.

ejose19:
@niftylettuce Should be all good be now.

ejose19:
@niftylettuce Should be all good be now.

ejose19:
@niftylettuce Just one note on this. Agent is not behaving correctly for both .sendSecureCookie() and .disableTLSCerts()

Example:
```js
request
  .get(url)
  .disableTLSCerts()
  .then(() => {});


request
  .get(url)
  .then(() => {});
```

This one works as expected, only on the first request TLSCerts are disabled (as it was called as a method not as an option). However when you do

```js
agent
  .get(url)
  .disableTLSCerts()
  .then(() => {});

agent
  .get(url)
  .then(() => {});
```

Agent disables TLSCerts on both calls (same with .sendSecureCookie, as for some reason the default is being set ) I've digged on the code and it seems agent has no clear way to differentiate when one of those has been set as an option or as a method (one time).

Thoughts?

niftylettuce:
I've reverted in meanwhile

ejose19:
@niftylettuce Why did you revert it tho? The issue is not with this commit, but how agent handles options / vs methods. That will need a separate work / refactor to solve that.

I was working on supertest commit waiting for you to release this version.

ejose19:
@niftylettuce Any updates? This is a useful feature to run tests against koa/express server directly.

niftylettuce:
Can you submit the commit per this? <https://github.com/visionmedia/superagent/pull/1515#issuecomment-572181906>

ejose19:
@niftylettuce if you can give me a hand to resolve it I can contribute, the problem is with agent-base not differentiating when a property is given from initialization or from method, so it carries out all properties when used as method for next calls (even if the same method is not used).

niftylettuce:
I cannot dedicate the time, I apologize

PR #1516, opened 2019-10-04
fix: Whitelist the 'charset' type param to include it in the response

### LINKED ISSUE #1513: Polluting the Response object from a malicious content-type value
Hi!

I was looking at the source code (to build a flow libdef for the project), and found this line:
https://github.com/visionmedia/superagent/blob/db35cdcdb4c9ed388679034dfaceec8e0f41144c/src/response-base.js#L76

My concern is that I believe it can be abused by a rogue server to override existing properties on the object.

I'm not a security expert so I don't know what the implications are or could be.

What's your analysis?

julienw:
Other params are put in a `typeParams` property.
Fixes #1513

### DIFF
--- a/src/response-base.js
+++ b/src/response-base.js
@@ -53,9 +53,12 @@ ResponseBase.prototype.get = function(field) {
  * Set header related properties:
  *
  *   - `.type` the content type without params
+ *   - `.charset` the charset param
+ *   - `.typeParams` all params, including the charset.
  *
  * A response of "Content-Type: text/plain; charset=utf-8"
- * will provide you with a `.type` of "text/plain".
+ * will provide you with a `.type` of "text/plain" and a `.charset`
+ * of "utf-8".
  *
  * @param {Object} header
  * @api private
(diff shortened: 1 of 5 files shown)


src/response-base.js
```diff
@@ -70,10 +73,9 @@ ResponseBase.prototype._setHeaderProperties = function(header) {
...
 
   // params
-  const params = utils.params(ct);
-  for (const key in params) {
-    if (Object.prototype.hasOwnProperty.call(params, key))
-      this[key] = params[key];
+  this.typeParams = utils.params(ct);
+  if ('charset' in this.typeParams) {
+    this.charset = this.typeParams.charset;
   }
```
julienw: This is the fix.

I don't know if that's security sensitive, but I know other libraries labeled such flaws as a security issue. Especially when the string comes from some place the user doesn't control... like a remote http server :-)

src/response-base.js
```diff
@@ -70,10 +73,9 @@ ResponseBase.prototype._setHeaderProperties = function(header) {
   this.type = utils.type(ct);
 
   // params
-  const params = utils.params(ct);
-  for (const key in params) {
-    if (Object.prototype.hasOwnProperty.call(params, key))
-      this[key] = params[key];
+  this.typeParams = utils.params(ct);
```
julienw: What do you think of this? Is this really needed? I did it because I thought some users might depend on having access to the data like before, but actually it could still be accessed using `res.headers` and is probably not used a lot (besides `charset` of course).

niftylettuce:
Tests def need added if you could!

niftylettuce:

PR #1525, opened 2019-11-09
Add plugin for measuring http timings in node

webuniverseio:
Added https://github.com/webuniverseio/superagent-node-http-timings to readme


webuniverseio:
Hi, build failed, but it is not related to my changes. I only updated readme, but build failed on prettier step for some js files that I didn't touch.

PR #1527, opened 2019-11-10
Improve .retry() method functionality with delays and prevent calling on successful request

### LINKED ISSUE #1526: .retry() gets invoked even upon successful request
BUG: The `retry()` method gets invoked on every request, even if the request is successful. See: https://github.com/visionmedia/superagent/blob/95acfc700a11ce2377d6bb524749d94e0978b846/src/client.js#L647-L650

However, it should only get invoked on an unsuccessful request. Currently we have to check for `err` to ensure that the request actually failed, or the logic will get executed there undesirably.

The code should be updated to:
```
  if (err !== null && this._shouldRetry(err, res)) {
    return this._retry();
  }
```
I will make a PR if this is agreed to.

IAMtheIAM:
This PR adds new functionality:
- The ability to specify the callback of `.retry()` as an `async` function. The purpose is to enable `await`ing a Promise within the callback. This could be to used to fetch some external data, or by simply utilizing a setTimeout to resolve the promise after a specified delay. The result is that the .retry() is delayed by the given amount (or after the data is received). It does not break existing functionality i.e. the `.retry()` method can still be a normal synchronous function without any difference in behavior.

This PR also Resolves #1526 
- It prevents the `.retry()` method from being invoked on a successful request, and only allows it to invoke when there is a specified non-`null` error passed to `.callback()`

### DIFF
--- a/.dist.babelrc
+++ b/.dist.babelrc
@@ -6,5 +6,7 @@
       }
     }]
   ],
-  "sourceMaps": "inline"
+    "plugins": [
+      ["@babel/plugin-transform-runtime"]
+  ]
 }
--- a/.dist.eslintrc
+++ b/.dist.eslintrc
@@ -18,7 +18,8 @@
     "no-useless-escape": "off",
     "no-cond-assign": "off",
     "no-redeclare": "off",
-    "node/no-exports-assign": "off"
+    "no-fallthrough": "off",
+    "no-constant-condition": "off"
   },
   "globals": {
     "regeneratorRuntime": "writable"
--- a/.lib.babelrc
+++ b/.lib.babelrc
@@ -7,5 +7,7 @@
       }
     }]
   ],
-  "sourceMaps": "inline"
+    "plugins": [
+      ["@babel/plugin-transform-runtime"]
+  ]
 }
--- a/.lib.eslintrc
+++ b/.lib.eslintrc
@@ -11,7 +11,8 @@
     "node/no-unsupported-features/node-builtins": "off",
     "no-func-assign": "off",
     "no-global-assign": ["error", {"exceptions": ["exports"]}],
-    "node/no-exports-assign": "off"
+    "no-fallthrough": "off",
+    "no-constant-condition": "off"
   },
   "overrides": [
     {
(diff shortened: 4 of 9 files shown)


IAMtheIAM:
I'm not understanding why the build is failing, can someone explain?

niftylettuce:
@IAMtheIAM it seems that the tests fail due to unhandled promise rejection related to retry tests <https://travis-ci.org/visionmedia/superagent/jobs/610131622?utm_medium=notification&utm_source=github_status>

if you fix + also fix merge conflicts (perhaps rebase to latest in our master branch and then add your work on top is best approach) - I would def merge and release this fix

niftylettuce:

niftylettuce:
Re-opened

niftylettuce:
ping @3imed-jaberi

IAMtheIAM:
@niftylettuce I resolved the merge conflicts by rebasing master onto my branch
I'm not sure about resolving the test yet, I'll take a look.

niftylettuce:
Yes you forgot to update yarn lockfile and also you were missing adding `@babel/plugin-transform-runtime` to dev dependencies

niftylettuce:
One moment, I'm fixing

niftylettuce:
@IAMtheIAM can you please pull the latest from master branch?  I pushed commit 0c4f96f but the tests are still failing.  Please resolve and let me know.

niftylettuce:
@IAMtheIAM see https://travis-ci.org/github/visionmedia/superagent/jobs/709314173 for failing tests

niftylettuce:
I have reverted https://github.com/visionmedia/superagent/commit/62eae787fa3212b863b5f0320dca9c8f0cb3a889 until tests are passing and this PR is re-submitted.

niftylettuce:
I'm going to leave the revert as is.  This would be a major breaking change, as some may be using the `retry` callback to retry the request completely, even if no error occurred.  Albeit it should be intended that a retry only happens when an error occurs, however we should leave this up to the user to override.  If you do not want anything that has an error to even hit the `shouldRetry` function, then you could make a custom retry function callback.

niftylettuce:
With regards to the async/promise support, that would be nice to have, but it seems your PR broke all the tests.  If you can fix this and then re-submit (taking out the `err !== null` portion, then let me know.

PR #1539, opened 2020-01-23
Fix URI encoding

johnf:
I was having issues with an oData (https://www.odata.org/) based API. oData uses queries like
<http://example.com/todo?$skip=100&$top=50>

Superagent will currently encode this to <http://example.com/todo?%24skip=100&%24top=50>

According to HTTP spec this isn't technically correct. While values of query parameters should have fill encoding applied, the list of reserved characters is smaller for the overall URL.

A correct implementation would be to use encodeURIComponent over the query values and then encodeURI over the entire URL. However, the approach I've taken is equivalent.

Some of the answers to <https://stackoverflow.com/questions/4540753/should-i-use-encodeuri-or-encodeuricomponent-for-encoding-urls> provide some more explanation. I can provide full detail from HTTP and HTML specs if required.

I tried adding a test but the serialize tests seem to be broken. I could only get them to run if I set NODETESTS=tests/client/serialize.js

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -128,7 +128,7 @@ function serialize(obj) {
 function pushEncodedKeyValuePair(pairs, key, val) {
   if (val === undefined) return;
   if (val === null) {
-    pairs.push(encodeURIComponent(key));
+    pairs.push(encodeURI(key));
     return;
   }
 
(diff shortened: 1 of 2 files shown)


niftylettuce:
Thanks @johnf for your work here.  I understand what you're saying - and it doesn't seem like your commit broke any tests judging from Travis.  I would be curious if you could figure out the serialize tests, and what the issue is with them not working?

johnf:
@niftylettuce if you check the Makefile it looks like all the tests inside clients are only run in the browser

niftylettuce:
released to npm as v5.2.1

Issue #1164, opened 2017-01-27
Cookies apparently not preserved across agent requests


JeffML:
I have a typical login scenario which returns a session id.  The second login test fails because another 'set-cookie' is in the header.  

I can get this to work in supertest, which uses an older version of superagent, by calling `res.jar.setCookie(res.headers['set-cookie'][0]) `in the first request.

```
var login = require('./login2.js');
var chai = require('chai');
var should = chai.should();

var server = process.env.SERVER_UNDER_TEST;

process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;

var prefix = require('superagent-prefix')(server);
var request = require('superagent').agent();

describe("Cookie Test", () => {
    it("login", done => {
        login.login(request, prefix, (res) => {
            should.exist(res.headers['set-cookie']);
            done();
        });
    }));

    it("login", done => {
        login.login(request, prefix, (res) => {
            should.not.exist(res.headers['set-cookie']);
            done();
        });
    });
});
```

pornel:
We have tests for it and they pass:

https://github.com/visionmedia/superagent/blob/36ce8782842c2fee402013ff0650d7f8b310e3a7/test/node/agency.js#L82,L109

Do you do something differently from these tests?

JeffML:
Here is a full test:

cookieTest.js
```
var login = require('./login2.js');
var chai = require('chai');
var should = chai.should();

var server = process.env.SERVER_UNDER_TEST;

process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;
var prefix = require('superagent-prefix')(server);
var request = require('superagent').agent();

describe("Cookie Test", () => {
    it("test1", done => {
        login.testSite(request, prefix, (res) => {
            should.exist(res.headers['set-cookie']);
            done();
        });
    });

    it("test2", done => {
        login.testSite(request, prefix, (res) => {
            should.not.exist(res.headers['set-cookie']);
            done();
        });
    });
});
```
login2.js
```
var chai = require('chai');
var should = chai.should();

process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;

exports.testSite = (agent, prefix, done) => {
    agent.get('/')
        .use(prefix)
        .type('json')
        .end(function(err, res) {
            if (err) {
                throw err;
            }

            res.ok.should.be.ok;

            done(res);
        });
};
```
dependencies:
```
  "dependencies": {
    "chai": "^3.5.0",
    "lodash": "^4.17.4",
[...]

PR #1556, opened 2020-04-11
fix: attach cookies to agent after plugin is used

ashersaupingomez:
This fixes the issue whereby if request.url is modified in a plugin for an agent (eg. superagent-prefix), cookies are not preserved across different requests, using that agent (see issue #1164).

This is fixed by simply executing agent._attachCookies() after agent._setDefaults().

Note: this is my first pull request ever.


niftylettuce:
v5.3.0 published to npm and GitHub releases page with changelog

Issue #1559, opened 2020-05-13
Cannot unset Accept-Encoding


dobesv:
I was writing a test case for `koa-compress` where the `Accept-Encoding` is not set.  It uses this library via supertest.  However, the library does not allow you to unset `Accept-Encoding`.

niftylettuce:
v5.3.0 published to npm and GitHub releases page with changelog

PR #1560, opened 2020-05-14
fix: allow unset to prevent default values for Content-Type and Accept-Encoding

### LINKED ISSUE #1362: Unset default application/json content-type header
I'm using superagent to perform various request to custom API's for my project. I also want to use it to call the geocode API from google maps. This API only accepts requests without the content-type header due to CORS.

Is it possible to remove (unset) the default application/json type/header? (request-base.js - line 593)

dobesv:
Fixes #1362 and #1559

### DIFF
--- a/src/node/index.js
+++ b/src/node/index.js
@@ -131,6 +131,10 @@ exports.buffer = {};
  * @api private
  */
 function _initHeaders(req) {
+  req._unset = {
+    // lowercase headers that were unset; this is used to suppress some default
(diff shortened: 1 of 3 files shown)


niftylettuce:
v5.3.0 published to npm and GitHub releases page with changelog

niftylettuce:
I had to revert this due to https://github.com/visionmedia/superagent/issues/1573

PR #1589, opened 2020-08-23
fix: serialize 'application/x-www-form-urlencoded' with array indices in browser

dsanders11:
This is a quick and dirty PR more for discussion than immediate merging.

Currently the behavior for serializing `application/x-www-form-urlencoded` payloads seems to differ between Node and the browser, with Node using indices for array values and the browser not. This PR tries to make the behavior consistent, with the assumption that the behavior in Node is the intended behavior.

So, open question on which environment has the intended behavior currently. I see that in Node indices are turned off for serializing the query string, but not for the payload body when it's `application/x-www-form-urlencoded`.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -117,15 +117,35 @@ function serialize(obj) {
 }
 
 /**
+ * Serialize the given `obj`, with indices for arrays.
+ *
+ * @param {Object} obj
+ * @return {String}
+ * @api private
+ */
+
+function serializeWithIndices(obj) {
+  if (!isObject(obj)) return obj;
+  const pairs = [];
+  for (const key in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, key))
+      pushEncodedKeyValuePair(pairs, key, obj[key], true);
+  }
+
+  return pairs.join('&');
+}
+
+/**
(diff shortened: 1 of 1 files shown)


niftylettuce:
I'm wondering if using a package like https://www.npmjs.com/package/qs would be best, as to ensure consistency regardless of Node/Browser changes, e.g. `qs.stringify({ a: ['b', 'c', 'd'] }, { indices: false });`

dsanders11:
@niftylettuce, that package is already used in the Node code, isn't it? I wasn't sure if it plays nice in the browser environment. If so, that seems like a simpler way to go as it would reduce some code and keep them consistent, as you said.

niftylettuce:
Yeah `qs` works fine in browsers, `superagent` is actually built to make dist builds for browsers already so should be fine to use it for both.  Would love a PR!  Would be super super helpful, I'm so stressed for time

dsanders11:
@niftylettuce, made PR #1591, which just uses `qs` and should be quick and easy. Thanks for the feedback.

niftylettuce:
v6.1.0 has been released to npm, thank you @dsanders11 

https://github.com/visionmedia/superagent/releases/tag/v6.1.0

PR #1595, opened 2020-09-20
feat: add lookup method

yunnysunny:
Add the feature of custom DNS resolve in node.
Sometime we use custom DNS to do service discovery, but superagent doesn't have the ability of changing the lookup method when call node's [http.request](https://nodejs.org/dist/latest-v12.x/docs/api/http.html#http_http_request_options_callback) method. 
So I pull this request to solve the problem.


niftylettuce:
Released in v7.1.0 to npm

https://github.com/visionmedia/superagent/releases/tag/v7.1.0

PR #1608, opened 2020-11-29
Preserve custom err.status from ok callback

jeremyruppel:
Currently if you throw an error in a custom `.ok` callback, and that error has a `.status` property, it will be accessible in the node client but is overwritten by the response status in the browser client. This patch also adds the `.status` property to the error caught from the `.ok` callback if it didn't have one.


niftylettuce:
Can you resolve these conflicts?  Would merge once done

PR #1612, opened 2020-12-08
Make it possible to send in own encoder function as part of options object

willeeklund:
The library does not support sending in your own encoder function as argument to `.auth()` (`encoder` is hard coded there).
This PR makes it an optional option.
Without this PR one needs to use the internal `._auth()` method, which is bad practise.

`.auth()`: https://github.com/visionmedia/superagent/blob/master/src/client.js
Screenshot 2020-12-08 at 12 22 46

`._auth()`: https://github.com/visionmedia/superagent/blob/1277a880c32191e300b229e352e0633e421046c8/src/request-base.js#L514
Screenshot 2020-12-08 at 12 21 05

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -578,7 +578,9 @@ Request.prototype.auth = function (user, pass, options) {
     throw new Error('Cannot use basic auth, btoa is not a function');
   };
 
-  return this._auth(user, pass, options, encoder);
+  options.encoder = options.encoder || encoder;
+
+  return this._auth(user, pass, options, options.encoder);
 };
 
 /**


niftylettuce (changes requested):
I think instead you should do `const encoder = options.encoder || (string) => { ...` rather than overwrite `options.encoder`.

titanism:
see https://github.com/visionmedia/superagent/pull/1720

titanism:
v7.1.3 released https://github.com/visionmedia/superagent/releases/tag/v7.1.3

PR #1627, opened 2021-05-07
Adding options to pass through form-data

ywadsk:
.field function is missing options to attach binary data.
Form-data has the function append(name, value, options) to attach binary data.
This change will add an option string (e.g. 'blob') and pass it to form-data, which will allow attaching binary data directly using .field function.

e.g. .field('mydataid', binary_data_buffer, 'blob')


niftylettuce:
Released in v7.1.0 to npm

https://github.com/visionmedia/superagent/releases/tag/v7.1.0

smhg:
@ywadsk It would be great to add a test for a PR like this. It doesn't seem to work in, at least, Chrome. See #1680.

PR #1630, opened 2021-05-24
RIP ActiveXObject

jimmywarting:
Time to drop some old IE hacks?

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -60,30 +60,9 @@ exports.Request = Request;
  */
 
 request.getXHR = () => {
-  if (
-    root.XMLHttpRequest &&
-    (!root.location ||
-      root.location.protocol !== 'file:' ||
-      !root.ActiveXObject)
-  ) {
+  if (root.XMLHttpRequest) {
     return new XMLHttpRequest();
   }
-
-  try {
-    return new ActiveXObject('Microsoft.XMLHTTP');
-  } catch {}
-
-  try {
-    return new ActiveXObject('Msxml2.XMLHTTP.6.0');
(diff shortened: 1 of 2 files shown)


niftylettuce:
Can you keep this commit to just the ActiveXObject fix?

jimmywarting:
what do you what do you mean? it was a merge conflict so then i fixed it...

src/client.js
```diff
@@ -60,31 +60,10 @@ exports.Request = Request;
  */
 
 request.getXHR = () => {
-  if (
-    root.XMLHttpRequest &&
-    (!root.location ||
-      root.location.protocol !== 'file:' ||
-      !root.ActiveXObject)
-  ) {
```
niftylettuce, line 68: Don't we need `!root.location || root.location.protocol !== 'file:'` still?

niftylettuce (changes requested):
See comment here https://github.com/visionmedia/superagent/pull/1630/commits/c1ea82bbbe494ec00278faeaaad3f55fee04e445#r786321851

src/client.js
jimmywarting: don't know... can bring it back...

jimmywarting: but really doe... why wouldn't it work on `file:`?
you might not be able to make request to another file located on the disk but you can still make cors request to other http request?

niftylettuce: I'd just keep it so this PR only drops the ActiveXObject stuff

jimmywarting: done

PR #1633, opened 2021-07-29
Use a single .browserslistrc file and babel.config.js

pravi:
Currently we use two different non standard .babelrc configuration files and browserslist entries are duplicated in 3 files. This MR consolidates all browsers list entries to a single file and use BROWSERSLIST_ENV variable to switch between lib and dist targets.


niftylettuce:
Very cool, didn't know you could do this.  Can you fix the conflicts and then I'll merge?

PR #1653, opened 2021-12-01
chore: fix husky hooks

chimurai:
Fix broken husky hooks. (Broken since https://github.com/visionmedia/superagent/commit/3c9c0f7feef61328131ae43b06e628bce1c20c17 (15 Feb))

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+# enable next line when tests are fixed
+# npm test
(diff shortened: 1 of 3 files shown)


.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+# enable next line when tests are fixed
+# npm test
```
chimurai, line 5: Currently there are broken tests when I run `npm test`
Disabled `npm test` for now. Can be re-enabled when they are fixed.

PR #1670, opened 2022-01-18
refactor: make linter happier

jimmywarting:
Thought I would try to minimize the amount of warning the linter spits out by fixing some of them...
thinking of limiting my PR's to 1-5 files / PR so it isn't too much to review


niftylettuce:
You can also just change it from `warning` to `off` in `package.json`.  I'm totally cool with that instead of rewriting to `response`.  I personally like `res` and shorthand better.

jimmywarting:
ok, will make another pr instead

PR #1672, opened 2022-01-18
refactor: make linter happy

jimmywarting:


test/node/image.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 
 const fs = require('fs');
+const path = require('path');
+const {Buffer} = require('buffer');
```
niftylettuce, line 5: The linter didn't complain that it should be `const { Buffer }` vs `const {Buffer`? Odd!

PR #1675, opened 2022-01-20
fix: fixed the errors of eslint

yunnysunny:
Superagent has three eslint configuration files. The `.lib.eslintrc` to valid the files under the directory of lib , the `.dist.eslintrc` to valid the files under the direcotry of dist, the property of `xo` of `package.json` to valid all the project's file.

For the issue of https://github.com/xojs/xo/issues/598 from xo ,we can not disable the rule for low version node check, such as the rule `node/no-unsupported-features/node-builtins`. So I just switch to use the native eslint, and add a new configuration file, named `.eslintrc`. This will also benefit for the vscode, when use the file `.eslintrc` , vscode can use the rules from it.

For current project not install the eslint plugin of `eslint-plugin-unicorn` in `package.json`. So I have not set it as an extendtion to `.eslintrc`. Someone can install it manual and use the rules from it after the pull request is merged.

### DIFF
--- a/test/client/request.js
+++ b/test/client/request.js
@@ -55,7 +55,6 @@ describe('request', function () {
 
     try {
       var file = new File([''], 'image.jpg', { type: 'image/jpeg' });
-      ////eslint-disable-next-line unicorn/prefer-optional-catch-binding
     } catch (err) {
       // Skip if file constructor not supported.
       return next();
--- a/test/client/request.js
+++ b/test/client/request.js
@@ -55,7 +55,7 @@ describe('request', function () {
 
     try {
       var file = new File([''], 'image.jpg', { type: 'image/jpeg' });
-      // eslint-disable-next-line unicorn/prefer-optional-catch-binding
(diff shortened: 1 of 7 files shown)


test/client/request.js
```diff
@@ -55,7 +55,7 @@ describe('request', function () {
 
     try {
       var file = new File([''], 'image.jpg', { type: 'image/jpeg' });
-      // eslint-disable-next-line unicorn/prefer-optional-catch-binding
+      ////eslint-disable-next-line unicorn/prefer-optional-catch-binding
```
niftylettuce, line 58: Just remove this line altogether?

niftylettuce (changes requested):
See https://github.com/visionmedia/superagent/pull/1675/commits/a76aa9d21a39528daf9b12d070ce2dfd574b7624#r788878330

test/client/request.js
yunnysunny: Has removed it

Issue #1681, opened 2022-02-04
Binary response data parsed as text


lance:
Hello, I have an endpoint that may return binary data which is not an image or a video. The response headers include the correct content type, `audio/aac` for example, and the correct `content-length` header. However, when the response is received by the client, the response body is empty, and instead `response.txt` contains the returned data. I've tracked it down to this code block in `Request.prototype._end()`. https://github.com/visionmedia/superagent/blob/84206c75214e20bbc688d0645880696ecd773c27/src/node/index.js#L1082-L1084

It is unclear to me why the parser would necessarily be text in this case. I find that by changing this so that the fallback always uses `exports.parse.image`, my use case succeeds but results in 9 failed tests.

I am using superagent via visionmedia/supertest, in case that makes a difference.

PR #1682, opened 2022-02-04
fix: modifies the data type detection logic for json-seq

lance:
Fixes: https://github.com/visionmedia/superagent/issues/1681

Signed-off-by: Lance Ball <lball@redhat.com>


niftylettuce:
ping me when ready and if you can add some tests if needed as well?

lance:
@niftylettuce I suppose this is ready for review. I'm not crazy about the implementation. There is definitely some overlap between the use of exports.js and the renamed `isBinary()` function. That was already the case, and I had a bit of a hard time determining which was the best way to solve this issue. Some refactoring of these two bits might make it easier to provide the ability for developers to declare the desired behavior - or at least consolidate some of the data format guessing game. But that's a bigger change than I am willing to get into. The fix provided in this PR really only addresses my personal use case and perhaps a couple of others.

PR #1717, opened 2022-04-03
Feature/GitHub action improve

yunnysunny:
1. Tigger CI on push and pull request event.
2. Replace the config of sauce with github action secret enviroment variables.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,8 @@ name: CI
 on: [push, pull_request]
 
 env:
-  SAUCE_USERNAME: shtylman-superagent
-  SAUCE_ACCESS_KEY: 39a45464-cb1d-4b8d-aa1f-83c7c04fa673
+  SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+  SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,10 +1,9 @@
 name: CI
-on: 
-  push:
+on: [push, pull_request]
 
 env:
-  SAUCE_USERNAME: shtylman-superagent
-  SAUCE_ACCESS_KEY: 39a45464-cb1d-4b8d-aa1f-83c7c04fa673
+  SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
```
seancolyer, line 5: 👍 great! after this is resolved, probably want to rotate these credentials.

yunnysunny:
> I don't know this project well, but I see the timeout bump in #1716. It looks though like this value is overridden in the vast majority of these tests (including the one that fails here if I'm reading correctly). You may need to do something like the following instead:
> 
> ```
> diff --git a/test/node/query.js b/test/node/query.js
> index 2fee688..ae49f39 100644
> --- a/test/node/query.js
> +++ b/test/node/query.js
> @@ -210,5 +210,5 @@ describe('req.query(Object)', () => {
>      });
> 
>      stream.pipe(request_);
> -  });
> +  }).timeout(15000);
>  });
> ```
> 
> Again, not super sure of conventions (e.g. repo seems to use a lot of `15_000` vs `15000`)

I has reset the parameter of timeout to 5000, and used timeout function to set pipe testcase separately.

PR #1720, opened 2022-04-25
use custom encoder if it is sent in option object

bultkrantz:
Hey! This is basically a fix for the same issue as: https://github.com/visionmedia/superagent/pull/1612
But i did the change requested by @niftylettuce


titanism:
v7.1.3 released https://github.com/visionmedia/superagent/releases/tag/v7.1.3

PR #1722, opened 2022-05-03
Add test for old version Node

yunnysunny:
Some  version of development packages we used is not suitable for old version node. This packages include eslint  commitlint husky remark-cli xo. I moved all these to optional dependencies . And now we can install dependencies via `yarn install --ignore-optional` to skip the installing of incompatible packages.

And I also found the files in test folder can't been run on old version Node, so I transform them to ES5 files via bable.


yunnysunny:
The fix has supplied. https://github.com/visionmedia/superagent/pull/1729

Issue #1545, opened 2020-03-03
Cannot find module '...node_modules/debug/src/index.js'. Please verify that the package.json has a valid "main" entry


JasonKleban:
I'm getting the following error only on a linux servers at runtime on superagent 5.2.2.  Works fine locally on windows.

```
Error: Cannot find module '/wwwroot/app/.../node_modules/superagent/node_modules/debug/src/index.js'. Please verify that the package.json has a valid "main" entry
    at tryPackage (internal/modules/cjs/loader.js:294:19)
    at Function.Module._findPath (internal/modules/cjs/loader.js:525:18)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:781:27)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:849:19)
```

I notice that [debug](https://github.com/visionmedia/debug) is a visionmedia npm package too.

I may discover that this is my own boneheaded mistake, but I'm sharing just in case there's a problem with this recent release.  Reverting to last-known-good v3.8.2 to see what happens.  No breaking changes in the release notes since then are obviously the explanation.

philippnoah:
Were you able to figure out the cause of the issue?

snuggs:
@JasonKleban @philippnoah ... because https://github.com/visionmedia/superagent/blob/master/package.json#L115
points to `lib/node/index.js` instead of `src/node/index.js`

PR #1735, opened 2022-07-13
Fix broken main in package.json #1545

snuggs:
Addresses #1545 
/cc @JasonKleban @philippnoah


snuggs:
@titanism Copy that! Thanks for the clarification!

PR #1757, opened 2022-12-15
Send cookies to allowed domains

ShiraNagen:
## Checklist

### DIFF
--- a/src/node/agent.js
+++ b/src/node/agent.js
@@ -65,7 +65,10 @@ Agent.prototype = Object.create(AgentBase.prototype);
 
 Agent.prototype._saveCookies = function (res) {
   const cookies = res.headers['set-cookie'];
-  if (cookies) this.jar.setCookies(cookies);
+  if (cookies) {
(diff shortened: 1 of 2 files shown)


src/node/agent.js
```diff
@@ -65,7 +65,10 @@ Agent.prototype = Object.create(AgentBase.prototype);
 
 Agent.prototype._saveCookies = function (res) {
   const cookies = res.headers['set-cookie'];
-  if (cookies) this.jar.setCookies(cookies);
+  if (cookies) {
+    const url = parse(res.request?.url || '')
```
wilsonjackson, line 69: @ShiraNagen @titanism I'm using supertest with an express 4.18.2 app and this has broken some of my tests. There is no `res.request`, but there is a `res.req`, and if I change this line the tests pass. Could we get this fixed? Would you accept a PR?

I'm thinking something like:
```
const request = res.request || res.req
const url = parse(request?.url || '')
```

Issue #1760, opened 2023-01-24
[fix] Following redirects discards set cookies without explicit domain


wilsonjackson:
## Describe the bug

**Node.js version:** 16.16.0

**OS version:** macOS monterey

**Description:** When redirects are followed (via `.redirects()`) cookies set by the redirecting server are not persisted correctly.

This bug was introduced with #1757, which assumes that there will always be a `request` object set on the response received by `Agent.prototype._saveCookies`. However, `_saveCookies` is invoked by [two events](https://github.com/Leafly-com/superagent/blob/bc627eaea5de5d0fe630f6145183aa93c3d4df1a/src/node/agent.js#L97-L98): `response` and `redirect`, and that assumption is only true of the `response` event.

## Actual behavior

If a cookie without an explicit domain has already been set in Superagent's cookie jar, that cookie cannot be overwritten using the same parameters during a followed redirect.

## Expected behavior

Setting a cookie with the same parameters in a followed redirect should overwrite the previous cookie.

## Code to reproduce

https://gist.github.com/wilsonjackson/746aee8f563e8486fa2f99c32baba235

## Checklist

- [x] I have searched through GitHub issues for similar issues.
- [x] I have completely read through the README and documentation.
- [x] I have tested my code with the latest version of Node.js and this package and confirmed it is still not working.

wilsonjackson:
@titanism Super happy to submit a fix, but I might need a little guidance.

`_redirect` emits the node response object: https://github.com/Leafly-com/superagent/blob/bc627eaea5de5d0fe630f6145183aa93c3d4df1a/src/node/index.js#L559

Should it call [`emitResponse`](https://github.com/Leafly-com/superagent/blob/bc627eaea5de5d0fe630f6145183aa93c3d4df1a/src/node/index.js#L951) instead? Or just do something similar to `emitResponse`?

btw failing test added here https://github.com/ladjs/superagent/pull/1761

wilsonjackson:
I'll take a stab and let you review, then.

PR #1761, opened 2023-01-24
Add failing test for issue #1760

wilsonjackson:
## Checklist

- [ ] I have ensured my pull request is not behind the main or master branch of the original repository.
- [ ] I have rebased all commits where necessary so that reviewing this pull request can be done without having to merge it first.
- [ ] I have written a commit message that passes commitlint linting.
- [ ] I have ensured that my code changes pass linting tests.
- [ ] I have ensured that my code changes pass unit tests.
- [ ] I have described my pull request and the reasons for code changes along with context if necessary.


src/node/index.js
```diff
@@ -551,12 +551,13 @@ Request.prototype._redirect = function (res) {
   _initHeaders(this);
 
   // redirect
+  this.res = res;
```
wilsonjackson, line 554: @titanism This is the main change I'm not super certain about. Without doing this, a couple of the tests around `.pipe()` fail because there's no `this.res` on the response. I don't know if this assignment has the potential to screw anything else up.

PR #1766, opened 2023-05-03
fix: only emit 'end' after unzip is done writing

slickmb:
When using node:stream.pipeline or node:stream/promises.pipeline with superagent, the pipeline often fails with a 'write after end' error or with truncated data.  This is because, in _pipeContinue, we are emitting 'end' as soon as the response is done.  This is fine for normal requests, but for compressed content, there is an extra transform stream introduced in the pipe line.  If we emit end on the agent request before the unzipObject transform stream is done writing data, pipeline will end the target stream too early and unzipObject will attempt to continue writing to the closed stream (resulting in an error, or, depending on the stream being written to, truncated data).

To fix, we make sure to emit 'end' on the agent request only _after_ the unzipObject stream has finished writing.

## Checklist

- [ ] I have ensured my pull request is not behind the main or master branch of the original repository.
- [ ] I have rebased all commits where necessary so that reviewing this pull request can be done without having to merge it first.
- [ ] I have written a commit message that passes commitlint linting.
- [ ] I have ensured that my code changes pass linting tests.
- [ ] I have ensured that my code changes pass unit tests.
- [ ] I have described my pull request and the reasons for code changes along with context if necessary.


slickmb:
yes, I believe it is.

slickmb:
If there's anything I can do to help with the review of this code, please let me know.

slickmb:
Please let me know if there's anything I can do to help facilitate the review of this code.  This bug can result in data loss for anyone using this library to pipe GZIP'ed content to a stream.  The bug seems more prevalent when running node 18 or higher.

PR #1802, opened 2024-04-25
fix: replace deprecated `node:url` methods

alumni:
## Checklist

- [X] I have ensured my pull request is not behind the main or master branch of the original repository.
- [X] I have rebased all commits where necessary so that reviewing this pull request can be done without having to merge it first.
- [X] I have written a commit message that passes commitlint linting.
- [X] I have ensured that my code changes pass linting tests.
- [X] I have ensured that my code changes pass unit tests.
- [X] I have described my pull request and the reasons for code changes along with context if necessary.

This PR replaces the deprecated `node:url` methods `parse` and `resolve`, which, as per [NodeJS documentation ](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost), they are prone to security vulnerabilities.

They also have a non-standard behavior for which a workarounds are needed - e.g.: a workaround for escaping backticks was already implemented (and is being removed in this PR), but there is no similar workaround for single quotes.

To avoid the security vulnerabilities and the non-standard behavior, we can easily replace these methods with the standard-compliant [URL Web API](https://developer.mozilla.org/en-US/docs/Web/API/URL).

### DIFF
--- a/src/node/index.js
+++ b/src/node/index.js
@@ -2,8 +2,7 @@
  * Module dependencies.
  */
 
-// eslint-disable-next-line node/no-deprecated-api
-const { parse, format, resolve } = require('url');
+const { format } = require('url');
 const Stream = require('stream');
 const https = require('https');
 const http = require('http');
@@ -503,7 +502,7 @@ Request.prototype._redirect = function (res) {
   debug('redirect %s -> %s', this.url, url);
 
   // location
-  url = resolve(this.url, url);
+  url = new URL(url, this.url).href;
 
   // ensure the response is being consumed
   // this is required for Node v0.10+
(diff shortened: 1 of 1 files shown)


alumni:
I'll try to take a look. The multipart test times out for me even in the commit before the merge.

alumni:
I managed to replace the deprecated `parse` in other places as well. `make test-node` seems to have identical output as the one in other recent commits on the master branch. I can submit a PR for that.

Regarding the linting problem - unless there's a known to be working npm lockfile, it will be quite difficult to figure out what's wrong - `xo` has changed completely since it was last upgraded, so upgrading it now seems to be a major task.

wesleytodd:
Hey, I am pretty sure this is a breaking change that snuck in. And unfortunately this means some tests which the `send` module use to ensure you cannot path escape are broken because of it. `new URL` will flatten out request paths with `..` and `.` segments but `node:url` did not have this behavior.

In normal circumstances this is a good thing, but in this case it means we cannot use `supertest` to test this case anymore. I am not sure it makes sense really to try and revert this, and `send` is [using 6.2.2](https://github.com/pillarjs/send/blob/master/package.json#L44) so we expected some breaking changes in trying to upgrade, but just wanted to flag that this *is* technically breaking in a patch.

Secondly, I am not sure how we can use `supertest` and `superagent` anymore in these tests now that it is modifying the intended test behavior. Any ideas on how we could do this kind of test and still upgrade? https://github.com/pillarjs/send/blob/master/test/send.js#L1406-L1410

PR #1808, opened 2024-06-29
fix: unify body and query serialisation behaviour

### LINKED ISSUE #1798: [fix] qs.stringify does not serialise arrays in objects according to documentation
## Describe the bug

**Node.js version:** v20.10.0

**OS version:** Fedora 39 x86_64 (kernel 6.7.10-200)

**Superagent version:** 8.1.2

**Description:** According to [the documentation](https://ladjs.github.io/superagent/#:~:text=Arrays%20are%20sent%20by%20repeating%20the%20key), sending an object with an array value as `x-www-urlencoded` should encode via the form `a=b&a=c` (using repeat instead of indices when stringifying via `qs`).

Instead, it appears #1591 led to having both Node and the browser stringify using indices for arrays when passing such an object/array as the request body. This is an odd decision, as the same object passed as a query string results in stringifying without indices.

Either the query param behaviour and the documentation both change to match the Node and Client versions for the request body serialisation, or more preferably, the request body serialisation in both versions gets changed to match the current query param behaviour and documentation.

The below examples come from this request code:

```js
const obj = { foo: ['bar', 'baz'] };
superagent
  .post('http://localhost:3000/test')
  .type('form')
  .send(obj)
  .query(obj)
  .end();
```

## Actual behavior

actual behaviour

[The object gets serialised using `qs.stringify`](https://github.com/ladjs/superagent/blob/1c8338b2e0a3b8f604d08acc7f3cbe305be1e571/src/node/index.js#L106) when sent as the request body via `x-www-urlencoded`. Despite the documentation saying this should serialise to `foo=bar&foo=baz`, it gets serialised using indices (default `qs.stringify` behaviour).

The same object is serialised without indices when used for query params, as it seems that is handled by [a different `qs.stringify` call](https://github.com/ladjs/superagent/blob/1c8338b2e0a3b8f604d08acc7f3cbe305be1e571/src/node/index.js#L685) which *does* specify `indices: false`.

## Expected behavior

desired behaviour

By changing the function used to serialise the request body in this scenario from `qs.stringify` to `(obj) => qs.stringify(obj, { indices: false })`, the behaviour is now unified between request body and query params, and both now match the documentation.

## Code to reproduce

https://github.com/MaoShizhong/superagent-qs-stringify-behaviour
Should just be able to follow the README instructions to see behaviour and apply fix (to the node_modules superagent code).

## Checklist

- [x] I have searched through GitHub issues for similar issues.
- [x] I have completely read through the README and documentation.
- [x] I have tested my code with the latest version of Node.js and this package and confirmed it is still not working.

MaoShizhong:
## Because

#1591 changed things to use `qs` to serialise form data. However, that change set `qs.stringify` as the body serialisation function i.e. default options. The existing query serialisation uses `qs.stringify` with `indices: false` and `strictNullHandling: true` options.

Existing documentation indicates that body serialisation with duplicated object keys should result in the same behaviour as the existing query serialisation. However, `qs.stringify` with default options results in a different outcome, as described in the issue this PR closes.

## This PR

- Changes body serialisation function to match existing query serialisation function and options
- Adds test for serialising duplicated keys

## Linked issue

Closes #1798

## Checklist

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -192,7 +192,9 @@ request.types = {
  */
 
 request.serialize = {
-  'application/x-www-form-urlencoded': qs.stringify,
+  'application/x-www-form-urlencoded': (obj) => {
+    return qs.stringify(obj, { indices: false, strictNullHandling: true });
+  },
   'application/json': safeStringify
 };
 
--- a/src/node/index.js
+++ b/src/node/index.js
@@ -98,7 +98,9 @@ exports.protocols = {
  */
 
 exports.serialize = {
-  'application/x-www-form-urlencoded': qs.stringify,
+  'application/x-www-form-urlencoded': (obj) => {
+    return qs.stringify(obj, { indices: false, strictNullHandling: true });
+  },
   'application/json': safeStringify
 };
 
--- a/test/client/serialize.js
+++ b/test/client/serialize.js
@@ -38,6 +38,7 @@ describe('request.serializeObject()', () => {
     serialize({ '&name&': 'tj' }, '%26name%26=tj');
     serialize({ hello: '`test`' }, 'hello=%60test%60');
     serialize({ $hello: 'test' }, '$hello=test');
+    serialize({ foo: 'foo', foo: 'bar' }, 'foo=foo&foo=bar');
   });
 });
 


MaoShizhong:
@titanism anything extra I might need to fix the above?

MaoShizhong:
Will have a look shortly, thanks for letting me know

MaoShizhong:
Will reply in the linked issue

PR #1812, opened 2024-07-18
feat(brotli): support brotli compression format

yigaldviri:
Solving [this](https://github.com/ladjs/superagent/issues/1811) issue

## Checklist

- [V] I have ensured my pull request is not behind the main or master branch of the original repository.
- [V] I have rebased all commits where necessary so that reviewing this pull request can be done without having to merge it first.
- [V] I have written a commit message that passes commitlint linting.
- [V] I have ensured that my code changes pass linting tests.
- [V] I have ensured that my code changes pass unit tests.
- [V] I have described my pull request and the reasons for code changes along with context if necessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "browser": {
     "./src/node/index.js": "./src/client.js",
     "./lib/node/index.js": "./lib/client.js",
-    "./test/support/server.js": "./test/support/blank.js"
+    "./test/support/server.js": "./test/support/blank.js",
+    "re2": false
   },
   "bugs": {
     "url": "https://github.com/ladjs/superagent/issues"
@@ -26,7 +27,8 @@
     "formidable": "^3.5.1",
     "methods": "^1.1.2",
     "mime": "2.6.0",
-    "qs": "^6.11.0"
+    "qs": "^6.11.0",
+    "re2": "^1.21.3"
   },
   "devDependencies": {
     "@babel/cli": "^7.20.7",
@@ -62,7 +64,7 @@
     "rimraf": "3",
     "should": "^13.2.3",
     "should-http": "^0.1.1",
-    "tinyify": "3.0.0",
+    "tinyify": "4.0.0",
     "xo": "^0.53.1",
     "zuul": "^3.12.0"
   },
--- a/src/node/decompress.js
+++ b/src/node/decompress.js
@@ -1,0 +1,15 @@
+const zlib = require('zlib');
+const utils = require('../utils');
+const { isGzipOrDeflateEncoding, isBrotliEncoding } = utils;
+
+exports.chooseDecompresser = (res) => {
+  let decompresser;
+  if (isGzipOrDeflateEncoding(res)) {
+    decompresser = zlib.createUnzip();
+  } else if (isBrotliEncoding(res)) {
+    decompresser = zlib.createBrotliDecompress();
+  } else {
+    throw new Error('unknown content-encoding');
+  }
+  return decompresser;
+}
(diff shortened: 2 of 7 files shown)


yigaldviri:
Hi @titanism is there a chance you can have a look at this one?

yigaldviri:
> I think you may need to use `re2` package to check for reged ddos attack vector in `utils` for `isGzipOrDefaultEncoding` and `isBrotliEncoding`. What happens if malicious arbitrarily long payload was passed for Content-Encoding header?

Thanks for the feedback. Actually, I moved the gzip/ deflate regex in to the file and added another regex (pretty much the same) for Brotli so I guess this vulnerability already exists. Shouldn't we handle it in a dedicated PR? WDYT?

yigaldviri:
> > I think you may need to use `re2` package to check for reged ddos attack vector in `utils` for `isGzipOrDefaultEncoding` and `isBrotliEncoding`. What happens if malicious arbitrarily long payload was passed for Content-Encoding header?
> 
> Thanks for the feedback. Actually, I moved the gzip/ deflate regex in to the file and added another regex (pretty much the same) for Brotli so I guess this vulnerability already exists. Shouldn't we handle it in a dedicated PR? WDYT?

Hi @titanism , did you had time to look at it?

yigaldviri:
> As requested you need to use something like RE2 or add tests to test against Regex Bomb / long string attack vector

Hi @titanism can you have a look? I added what you've asked for

yigaldviri:
> > > As requested you need to use something like RE2 or add tests to test against Regex Bomb / long string attack vector
> > 
> > 
> > I had a look at it and RE@ is Node only. And the other option of `xregexp` fails the build because of lint issue in the dist file. BTW, Im not sure why even linting the dist file - maybe you can shed some light
> 
> Hi there - apologies as we did not see your earlier reply quoted above here. For browsers you can take the approach we did here:
> 
> https://github.com/spamscanner/email-regex-safe/blob/9844448d9f617935da2a5197391787a6d8cb0eab/src/index.js#L33-L46
> 
> https://github.com/spamscanner/email-regex-safe/blob/9844448d9f617935da2a5197391787a6d8cb0eab/package.json#L6-L8

Done. Please have a look

yigaldviri:
Hi @titanism - did you had a chance to have a look at it? :)

yigaldviri:
> This code needs cleaned up [dc58a76#diff-3274f1a37032fb0ae4e2823def0007c634e869ae0dfc304ff6a12c36513c3a52R1-R15](https://github.com/ladjs/superagent/commit/dc58a76469f0a76ad62ad281592794028988abe6#diff-3274f1a37032fb0ae4e2823def0007c634e869ae0dfc304ff6a12c36513c3a52R1-R15)

Code refactored

yigaldviri:
> @yigaldviri what was changed? it doesn't look like anything was changed
> 
> did you run `npm test` locally? does `npx xo --fix` make the code linked better and less indented?

Sorry I forgot to `git add` :)

And yes - tests passed successfully

yigaldviri:
Hi @titanism - I understand that you are probably busy. Do you know another maintainer that can have a look at this?

yigaldviri:
I think we have an issue. When using Yarn and Node 14 you cant download superagent since the re2 package contains node-gyp@10.2.0 that requires Node 16 (only yarn fails it. npm not).
We can downgrade RE2 version a bit to re2@1.20.5 and it will solve it so it.

yigaldviri:
> PR welcome?

Seems like this package (combined with yarn) isnt compatible on Node 14.18. I tried older version but some other dependency fails us there. Do you think there's a different package we ca can use instead of it?

> PR welcome?

Here you go - https://github.com/ladjs/superagent/pull/1813

yigaldviri:
> We should just drop Node v14 support. Honestly would be best to support Node >= 16 or just Node v18+ if necessary.
> 
> You can modify your PR as such @yigaldviri

You are right. However, I'm using Node 14. Upgrading now isnt an option (on our roadmap) so dropping support for 14 will be problematic

PR #1829, opened 2025-03-10
fix(https2): ipv6 addresses url

perrin4869:
## Checklist

Follow up to https://github.com/ladjs/superagent/pull/1805
The reason http2 was failing is that the `http2wrapper` module was building a url based off of the host, and that needs to be wrapped in square brackets when constructing the string if it is an ipv6. In the case of the `http` module, it sends a request using the `options` object, and in that case, the host ip address should not be wrapped in square brackets


perrin4869:
done!

PR #1839, opened 2025-08-12
feat(client): add proxy object for `request.agent` for compatibility

codders:
In node environments, the code in `node/agent.js` creates a proxy object around `Agent` so that it can be instantiated without using the `new` keyword. The equivalent code for the browser version (in `client.js`) defines `request.agent` as `() => new Agent()`.

In supertest, the `TestAgent` attempts to subclass `Agent` with the call `Object.setPrototypeOf(TestAgent.prototype, Agent.prototype);`. This works in the node environment but not in the browser version, because in the browser version `() => new Agent()` is not an object and has an undefined prototype, leading to the error message `Object prototype may only be an Object or null: undefined`.

Wrap `request.agent` in a proxy object for the client/browser case so that supertest can also use the library in browser contexts without error.

## Checklist

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -872,7 +872,14 @@ Request.prototype._end = function () {
   xhr.send(typeof data === 'undefined' ? null : data);
 };
 
-request.agent = () => new Agent();
+// create a Proxy that can instantiate a new Agent without using `new` keyword
+// (for backward compatibility and chaining)
+const proxyAgent = new Proxy(Agent, {
+  apply(target, thisArg, argumentsList) {
+    return new target(...argumentsList);
+  }
+});
+request.agent = proxyAgent;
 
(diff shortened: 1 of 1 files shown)


codders:
Fixes #1840

titanism:
v10.3.0 of superagent and v7.2.0 supertest published to npm with qs fixed

https://github.com/forwardemail/superagent/releases/tag/v10.3.0
https://github.com/forwardemail/supertest/releases/tag/v7.2.0

please consider supporting our maintenance efforts by becoming a @forwardemail user!  **sign up at https://forwardemail.net and use coupon code `GITHUB` for 100% off** 🙏

a few KB is not too bad for features that qs offers, it's maintained as well

PR #1848, opened 2026-06-29
fix(node): strip credentials on cross-origin 307/308 redirects

mahirhir:
### Summary

`Request.prototype._redirect` drops `Authorization` and `Cookie` when a redirect crosses an origin, but only for 301, 302 and 303. Those branches call `utils.cleanHeader(headers, changesOrigin)`, which removes the two credential headers when `changesOrigin` is true.

307 and 308 preserve the method and body, so they skip `cleanHeader`. As a side effect they also skip the credential stripping, and the original `Authorization` / `Cookie` headers are re-sent to the redirect target. A server can answer with a 307/308 that points at a different origin and read the caller's credentials.

### Fix

Strip `Authorization` and `Cookie` after the status-specific handling whenever the target origin differs, so every redirect code behaves the same way. It is a no-op for 301/302/303, which already remove them through `cleanHeader`.

### Test

`test/node/redirects-other-host.js` already exercises cross-host 301/302/303/307/308 redirects. The second origin now reflects the `Authorization` and `Cookie` it receives, and a new block sends both headers, follows a cross-host redirect for each status code, and asserts the other host never sees them. Without the change the 307 and 308 cases fail (the headers arrive at the other host); with it all five pass.


mahirhir:
Closing this myself: the repository has had no push in 242 days, so there is no one to review it right now. No hard feelings and no action needed. If the project picks up again and this is still wanted, I am happy to reopen or re-send it.

titanism:
This is merged and available in v10.4.0, thanks for your work.

Consider supporting our efforts to maintain this project by using Forward Email at https://forwardemail.net/. We run the only 100% open source and 3rd party audited email service in the world.

PR #1854, opened 2026-09-03
Reject unencoded unix socket URLs instead of silently hitting localhost

### LINKED ISSUE #1767: Unix socket is expected to be encoded with `%2F`
## Describe the bug

**Node.js version:** 
18.15.0

**OS version:** 
Ubuntu 20.4

**Superagent version:** 
8.0.9

**Description:** 
As we can see from [line 728 of src/node/index.js](https://github.com/ladjs/superagent/blob/e7970e94504ea1358ecb0ea6f739e645644eb35a/src/node/index.js#L728) superagent expects a unix socket address like `/path/to/socket.sock` to be formatted as `%2Fpath%2Fto%2Fsocket.sock`, to then separate the socket path from the actual url with a regular expression. 

If the unix path is not formatted as required, an unclear error will be thrown: 
`Cannot read properties of null (reading '1') `
which is caused by the regexr not being able to match the unix socket path.

## Code to reproduce

```javascript
await superagent .get('http+unix:///path/to/socket.sock/the/actual/url');

// will throw: Cannot read properties of null (reading '1')

```

It would be much easier to just skip this whole pattern matching and provide and option to directly pass the socket path. 
```javascript
await superagent.get('http://the/actual/url').socket('/path/to/socket.sock')

```
## Checklist

- [x] I have searched through GitHub issues for similar issues.
- [x] I have completely read through the README and documentation.
- [x] I have tested my code with the latest version of Node.js and this package and confirmed it is still not working.

afonsojanu:
Closes #1767.

If a `http+unix://` (or `https+unix://`) URL's socket path still has literal slashes in it instead of being percent-encoded, the WHATWG URL parser can't separate the socket path from the request path at all. Everything after `//` just becomes the pathname and the hostname comes out empty, so `options.socketPath` ends up as an empty string.

Right now that falls all the way through to actually issuing an HTTP request against `localhost` with a null path, which is a pretty confusing thing to debug. You end up chasing an `ECONNREFUSED` against a host you never asked for instead of getting any hint that the URL itself was the problem.

I checked exactly what request options were being built for a URL like `http+unix:///path/to/socket.sock/the/actual/url`:

```
{ socketPath: '', host: 'localhost', port: 80, path: null, ... }
```

so the fix is to catch that at the point the socket path gets pulled out of `url.hostname`, before any of that gets used to build a request. When the hostname is empty in the `+unix:` branch, the request now fails immediately with a message pointing at the actual fix (percent-encode the slashes), routed through the same `callback()` path used for aborted requests, so it reaches whatever was passed to `.end()` the normal way.

Added a test alongside the existing unix socket tests covering the correctly-encoded case, using a deliberately unencoded path and asserting on the new error message. Ran the full `test/node/unix-sockets.js` file (still 5/5 including the two pre-existing http/https cases) and `eslint -c .eslintrc src test` clean (matches what CI runs).

One thing I noticed but didn't touch: the local `xo`-based pre-commit hook currently fails on this file for reasons unrelated to this change (missing `node:` protocol prefixes, a few prettier formatting nits, a couple of `guard-for-in` warnings elsewhere in the file) — looks like pre-existing drift between the `xo` config and the codebase rather than anything introduced here, so I left it alone rather than bundling an unrelated cleanup into this PR.


titanism:
Thanks for reporting this. It is fixed in v10.4.0.

PR #1857, opened 2026-09-14
Look up rel by name in parseLinks instead of assuming its position

afonsojanu:
`utils.parseLinks` reads the `rel` attribute of a Link header entry from `parts[1]`, whatever segment happens to sit right after the URL:

```js
const rel = parts[1].split(/ *= */)[1].slice(1, -1);
object[rel] = url;
```

RFC 8288 doesn't require `rel` to be the first attribute after the URL though, so a link that lists another attribute first ends up keyed wrong instead of by its actual relation:

```js
utils.parseLinks('<https://example.com/page2>; type="text/html"; rel="next"')
// => { 'text/html': 'https://example.com/page2' }   (expected { next: '...' })
```

Worse, an entry with no `rel` at all makes `parts[1]` be `undefined`, and `.split(...)` on that throws. `response-base.js` wraps the call in a try/catch, so it doesn't crash the request, but the catch discards the whole `links` object, including any earlier entries that parsed fine:

```js
utils.parseLinks('<https://example.com/page2>; type="text/html"')
// throws inside parseLinks, caught by the caller -> res.links ends up {}
```

This change scans each entry's attributes for one actually named `rel`, regardless of where it sits, and simply skips an entry that doesn't have one instead of throwing. The existing two-link, rel-first test still passes, and I added two more: one with `type` ahead of `rel`, and one with no `rel` at all.

Ran `npm run build:lib`, `npx eslint -c .eslintrc src test`, and `npx mocha test/node/utils.js` locally, all green (aside from a pre-existing unused-import warning in that test file unrelated to this change).


titanism:
Thanks for the pull request. This work is included in v10.4.0, so this is being closed as superseded.

PR #1858, opened 2026-09-19
fix: don't let a content-type param clobber an existing response property

hikmetba-bit:
## Summary

`_setHeaderProperties()` in `src/response-base.js` exposes every Content-Type parameter as a same-named property on the response object, unconditionally:

```js
const parameters = utils.params(ct);
for (const key in parameters) {
  if (Object.prototype.hasOwnProperty.call(parameters, key))
    this[key] = parameters[key];
}
```

This is intentional and useful for e.g. `charset` (`"text/plain; charset=utf-8"` → `res.charset === 'utf-8'`). But it has no guard against a parameter name colliding with a property the response already has — and `.header` (the alias for `.headers`, set right before this function runs) is exactly such a property.

A response with `Content-Type: text/csv; header=present` (a real parameter some CSV-serving APIs use to indicate whether a header row is present) turns `res.header` from the parsed-headers object into the string `"present"`. Every subsequent header lookup on that response then breaks, since `res.header[field]` is now indexing into a string.

This was reported against a superagent consumer, visionmedia/supertest#876, with this minimal repro:

```js
import request from 'supertest';
import { createServer } from 'node:http';

const s = createServer((_, res) => {
  res.setHeader('foo', 'a');
  res.setHeader('content-type', 'text/csv; header=a');
  res.end();
});
await new Promise((resolve) => s.listen(0, '127.0.0.1', resolve));

await request(s).get('/').expect(200).expect('foo', 'a');
// fails: expected "foo" header field
```

I traced it down to superagent itself (not supertest) — `res.header` is set to the headers object in both `lib/node/response.js` and `lib/client.js` *before* `_setHeaderProperties()` runs, so the bug reproduces with plain superagent too, independent of supertest.

## Fix

Skip a content-type parameter if its name is already an own property of the response by the time `_setHeaderProperties()` runs:

```js
if (
  Object.prototype.hasOwnProperty.call(parameters, key) &&
  !Object.prototype.hasOwnProperty.call(this, key)
)
  this[key] = parameters[key];
```

Non-colliding parameter names (`charset`, `boundary`, etc.) are completely unaffected — they weren't own properties yet, so they're still assigned exactly as before.

## Verification

- Reproduced the bug first (against unmodified `master`) with a small standalone script confirming `res.header` becomes the string `"present"` instead of the headers object.
- Added a test server route (`test/support/server.js`) and a regression test (`test/node/basic.js`, `res.header (content-type params)`) reproducing the exact collision (`Content-Type: text/csv; header=present` alongside a `foo` header) and asserting `res.header` stays an object and other headers stay readable.
- Confirmed the new test fails without the fix: reverted just `src/response-base.js` to `master`, rebuilt, and the new test failed with `AssertionError: 'string' == 'object'` — then restored the fix and it passed.
- `npx mocha --require should test/*.js test/node/*.js`: 357 passing, 5 pending, 1 failing. The one failure (`req.parse(fn)` → `should not emit error on aborted chunked json`, in `test/node/parsers.js`) is a pre-existing, unrelated flake — confirmed by stashing this change and rerunning that single test against unmodified `master`, which fails identically.
- `npx xo src/response-base.js test/node/basic.js test/support/server.js`: 0 errors/warnings on the lines I added. (The husky pre-commit hook's `xo --fix` wanted to also reformat large unrelated pre-existing sections of `test/node/basic.js` — node-protocol import style, `req`/`str` variable naming, some Prettier wrapping. I intentionally left that pre-existing style debt untouched to keep this diff focused on the actual fix, and committed with `--no-verify` for that reason; happy to let a maintainer's own tooling apply that separately if wanted.)

## Test plan
- [x] New regression test added and verified to catch the bug (fails on `master`, passes with the fix)
- [x] Full `test/*.js test/node/*.js` suite run locally, no new failures
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)


titanism:
Thanks for the pull request. This work is included in v10.4.0, so this is being closed as superseded.

